PR #94, opened 2015-01-09
Respect process' umask when creating new files

flaviogrossi:
When new files are created, the file permissions should respect the process umask, to give some permission control to the user of programs which use libsndfile.


erikd:
Applied manually. Thanks.

PR #100, opened 2015-04-22
Fix memory leaks

DanWin:


erikd:
Applied manually. Thanks!

PR #121, opened 2016-02-02
use svg travis badge

MartinDelille:


erikd:
Is there any advantage to the SVG travis badge? Sorry, I don't understand why this is even being proposed.

MartinDelille:
SVG badges are light weight and display better on some screen (like retina).

PR #125, opened 2016-03-13
Separate out the configure flags for libflac, libogg and libvorbis

martinwguy:
Previously you could only configure all of libflac, libogg and libvorbis
or none of them. This change separates out the detection/enabling/disabling
of them into --enable-libflac --enable-libogg and --enable-vorbis.

Apart from being useful in itself, this makes it possible to
add other formats via external libraries (libopus, libspeex ...)
without insisting that all of the external libraries be available or none.

We keep the old option --disable-external-libs because it is used in the
build scripts for OpenWRT, OpenEmbedded and others.

See the discussion in https://github.com/erikd/libsndfile/issues/15

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -126,8 +126,19 @@ AC_ARG_ENABLE(sqlite,
 AC_ARG_ENABLE(alsa,
 	AC_HELP_STRING([--disable-alsa], [disable use of ALSA]))
 
+AC_ARG_ENABLE(libflac,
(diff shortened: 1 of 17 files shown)


martinwguy:
Not required. erik wants all three under a single option.

erikd:
Not required *now* and not required to add Wavpack. The decision can be revisited when Opus or Speex is added.

PR #128, opened 2016-04-01
Updated cue feature patch

larspalo:
Here's an updated patch that add cue/mark feature for both aiff and wav.


erikd:
This has been merged and tweaked.

Thanks, this was a major contribution.

PR #139, opened 2016-08-21
Working CMake project (Win32, Linux, MacOS)

evpobr:
# Summary

It passes Travis CI tests under Linux and MacOS.

Also tested to compile under Visual Studio 2015 with Clang compiler toolset (VS 2015 update 1 supports it).

# Changes

* Compiltely reworked CMakeLists.txt;
* `sf_unistd.h`: `R_OK`, `W_OK`, `X_OK `and `F_OK `are defined if they are missing;
* some source files now use `sf_unistd.h` if `unistd.h` missing.


erikd:
Whats the difference between the debug and release builds?

evpobr:
@erikd 

CMAKE_C_FLAGS_RELEASE: `-O3 -NDEBUG`

CMAKE_C_FLAGS_DEBUG: `-g`

CMakeLists.txt
```diff
@@ -5,24 +5,24 @@
 # lowest version supporting C_STANDARD property
 cmake_minimum_required (VERSION 3.1.3)
 
+# Set software information
+project(libsndfile VERSION 1.0.28.0)
+
```
erikd: Why does that have the `.0` on the end. I have always used an `x.y.x` version number system for libsndfile.

evpobr: I wasn't sure about that.

erikd: I think the final `.0` may be needed for something in the Windows build chain, but should not show up in the non-Windows builds.

evpobr:
@erikd If you never use the `X.X.X.X` version, I can replace it with `X.X.X`.

The last zero from `PROJECT_VERSION` (aka `PROJECT_VERSION_TWEAK`) is used in the [version-metadata.rc.in.cmake](/evpobr/libsndfile/blob/fix-package-version/src/version-metadata.rc.in.cmake#L5). But it can be simply replaced by zero in place.

CMakeLists.txt
evpobr: If you never use the `X.X.X.X` version, I can replace it with `X.X.X`.

The last zero from `PROJECT_VERSION` (aka `PROJECT_VERSION_TWEAK`) is used in the [version-metadata.rc.in.cmake](/evpobr/libsndfile/blob/fix-package-version/src/version-metadata.rc.in.cmake#L5). But it can be simply replaced by zero in place.

erikd: Sounds good! Make it so.

evpobr:
@erikd, can you create experimental branch and merge all the CMake related work? I want to delete my fork from @syb0rg and make fork from you directly.

erikd:
Will do this some time this weekend.

erikd:
This PR has been merged onto the `cmake-config` branch.

PR #141, opened 2016-09-02
Autotools: LT_INIT

evpobr:
`AM_PROG_LIBTOOL` is [obsolete](https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html), use `LT_INIT` instead.


erikd:
Thanks!

PR #142, opened 2016-09-03
configure.ac: improve 'ssize_t` checks

evpobr:
Now we typedef `ssize_t` to appropriate type, if it is missing.


erikd:
The PR comment don't actually match the code. Is that want you meant?

evpobr:
@erikd, are you talking about "Remove unneeded check"? Hmm, it's my fault, result of my bad skills in commits squashing.

Commited changes are OK. `AC_TYPE_SSIZE_T` is recommended way to test `ssize_t`. If `ssize_t` is not present, it will be declared in `config.h` as:

    typedef int ssize_t;

PR #145, opened 2016-09-07
Remove truncate calls

evpobr:
Windows lacks `truncate`, use more cross-platform solution ([`fopen` truncates existing file in write mode](http://pubs.opengroup.org/onlinepubs/009695399/functions/fopen.html)).


erikd:
Good idea, but I came up with a slightly neater solution.

PR #146, opened 2016-09-07
Ignore config.h.in~ file

evpobr:
config.h.in~ is autogenerated file, no need to track it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,6 +89,7 @@ src/Symbols.os2
 src/Symbols.static
 src/config.h
 src/config.h.in
+src/config.h.in~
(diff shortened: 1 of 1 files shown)


erikd:
Would prefer this on the `cmake-config` branch as well.

When the CMake config stuff is complete, I'd like to squash all the commits and it make sense to add things like this to the big squashed commit rather than adding to the master branch now.

evpobr:
May be just rebase this branch on each new commit to master? Both branches will be synchronized, with CMake changes only here.

erikd:
Merged into `cmake-config` branch.

PR #147, opened 2016-09-07
Ignore INSTALL file

evpobr:
INSTALL is autogenerated file, no need to track it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,4 @@ tests/test_wrapper.sh
 tests/utils.c
 tests/utils.h
 tests/write_read_test.c
+INSTALL


erikd:
`INSTALL` not an autogenerated file. It was added in commit:
```
commit cb979e10a703033864f8f42c94e9d1d335e5be40
Date:   Mon May 14 19:55:24 2007 +1000

    First snapshot of the public project.
```
and has not changed since.

evpobr:
`INSTALL` is not autogenerated, my mistake. It is just a symlink to the file [somewhere deep inside of automake directory](https://groups.google.com/forum/m/#!topic/libpsl-bugs/v0KhOXgVhEY). And autogen.sh or autoreconf -i will restore it back.

erikd:
`autoreconf -i` will only recreate it if you delete it. If the file is already there, `audoreconf` won't touch it.

evpobr:
I mean why keep this file in Git index, if it is always can be created during configuration? We don't keep `configure`, `config.h`, `Makefile.in's`, `makefile's` and etc, so why keep `INSTALL`? It has some manually provided information?

erikd:
The intention of the automake `INSTALL` is to provide a basis for a project specific `INSTALL` file.

Once we have proper CMake support, this `INSTALL` file should be updated to include info about installing via CMake.

evpobr:
> The INSTALL file: Because the GNU installation procedure is streamlined, a standard `INSTALL' file will be created for you automatically by Automake. If you have something very important to say, it may be best to say it in the `README' file instead. the `INSTALL' file is mostly for the benefit of people who've never installed a GNU package before. However, if your package is very unusual, you may decide that it is best to modify the standard INSTALL file or write your own.

It seems `INSTALL` file isn't good place for CMake information.

I found it [here](https://www.st-andrews.ac.uk/~iam/docs/tutorial.html).

Also, `autoreconf -ivf` rewrites this file completely.

PR #149, opened 2016-09-09
Remove unneeded branch of code

evpobr:
With CMake we can configure sndfile.h properly.

### DIFF
--- a/src/sndfile.h.in
+++ b/src/sndfile.h.in
@@ -329,17 +329,10 @@ typedef	struct SNDFILE_tag	SNDFILE ;
 /* The following typedef is system specific and is defined when libsndfile is
 ** compiled. sf_count_t will be a 64 bit value when the underlying OS allows
 ** 64 bit file offsets.
-** On windows, we need to allow the same header file to be compiler by both GCC
-** and the Microsoft compiler.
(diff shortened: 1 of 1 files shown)


erikd:
This could break the autotool configuration. I'm not ready to replace autotools with CMake. For the time being at least libsndfile will have two configuration methods.

evpobr:
Can you explain please? This build passes Travis CI check with Autotools.

erikd:
I sure it does build with that in Travis because it is a Windows specific hack.

When configured for Windows, but built with MinGW:
```
#if (defined (_MSCVER) || defined (_MSC_VER))
typedef __int64         sf_count_t ;
#define SF_COUNT_MAX            0x7fffffffffffffffi64
#else
typedef __int64 sf_count_t ;
#define SF_COUNT_MAX            0x7FFFFFFFFFFFFFFFLL
#endif
```
The first branch is needed because MSVC does not understand the `LL` type modifier.

evpobr:
Hmm, VS2015 understands `LL`.

erikd:
Earlier versions may not.

evpobr:
You're right, i close this PR.

PR #150, opened 2016-09-10
Use io.h as partial replacement of unistd.h

evpobr:
This fixes some aspects of missing `unistd.h` header in  MSVC.
`sf_unistd.h` defines some useful constants, now it also includes
`io.h`, where MSVC declares [required](https://github.com/erikd/libsndfile/blob/master/src/common.c#L1702) `access` function.


erikd:
Can't this commit be added to the cmake hacking branch?

evpobr:
If you wish.

PR #152, opened 2016-09-10
Use io.h header if unistd.h is missing, closes #151

### LINKED ISSUE #151: Missing unistd.h header under Win32
We are using POSIX-functions (e.g. `access`), usually declared in `unistd.h` header, but MSVC doesn't have it. Some POSIX-functions are declared in `io.h`, so we must detect and include it under Windows.

evpobr:


erikd:
Merged into `cmake-config` branch.

Issue #155, opened 2016-09-11
Fix MSVC warning about SF_COUNT_MAX macro redifinition


evpobr:
Defined in sndfile.h:

    #define SF_COUNT_MAX 0x7fffffffffffffffi64

In config.h:

    #define SF_COUNT_MAX 0x7FFFFFFFFFFFFFFFLL

Different definitions causes this warning.

This bug is is related to #149.

PR #156, opened 2016-09-12
Properly select MSVC version when define SF_COUNT_MAX in sndfile.h #155

evpobr:
Miscosoft C compiler supports `long long` and `LL` 64-bit integer-suffix [since Visual Studio 2003](https://msdn.microsoft.com/en-us/library/00a1awxf(v=vs.71).aspx).

`_MSCVER` macro is dropped because i didn't find any information about it in [MSDN](https://msdn.microsoft.com/en-us/library/b0084kay.aspx).

Mybe it's better just do not support this MSVC-specific code, these compiler versions are really old.

I think it makes no sense to add related detection code in CMake, due to lack of normal support of C in MSVC before 2013, which supports `long long` and `LL`.

### DIFF
--- a/src/sndfile.h.in
+++ b/src/sndfile.h.in
@@ -333,7 +333,7 @@ typedef	struct SNDFILE_tag	SNDFILE ;
 ** and the Microsoft compiler.
 */
 
-#if (defined (_MSCVER) || defined (_MSC_VER))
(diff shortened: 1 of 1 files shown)


erikd:
Merged!

Issue #157, opened 2016-09-12
Improve C inline support


evpobr:
Some related checks exists to enable inline support in GCC, but nothing is done to to support `__inline__` or `__inline`. I guess MS supports `inline` from VS2015 only, but has `__inline`.

erikd:
`common.h` could have something like
```
#if defined _MSVC && ! defined (inline)
#define inline __inline
#endif
```

evpobr:
VS2013 build failed - it doesn't know what is `inline`. Yes, we detect and configure `inline` properly, but some code do not use global `config.h`. It is `libgsm` and `libalac`. `libg72x` is affected too, but it doesn't use `inline`.

erikd:
Would it be possible to just say VS2013 is not supported?

evpobr:
I thought about it. It is better to fix this. Any compiler in C90-compatibility mode will suffer.

erikd:
Absolutely. C99 is a 16+ year old standard. Its more than OK to require it.

PR #158, opened 2016-09-12
configure.ac: check for inline #157

evpobr:
If the C compiler supports the keyword inline, do nothing. Otherwise
define inline to __inline__ or __inline if it accepts one of those,
otherwise define inline to be empty.


erikd:
Merged this to the `cmake-config` branch.

evpobr:
@erikd, i don't see merge.

erikd:
I merged but was waiting for travis to finish testing on the `cmake-config` branch. It should be there now.

PR #159, opened 2016-09-12
CMake: check for inline, closes #157

### LINKED ISSUE #157: Improve C inline support

evpobr:
Similar to #158, but for CMake.


erikd:
Merged. Thanks.

PR #163, opened 2016-09-13
sndfile-salvage.c: add missing headers

### LINKED ISSUE #162: MSVC build: sndfile-salvage fails because of missing unistd.h
Here is compiler output:

    sndfile-salvage.c(40): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory

evpobr:
Add sfconfig.h, under MSVC use sf_unistd.h.


evpobr:
Should be in master branch.

PR #165, opened 2016-09-13
MSVC: fix stack overlow in floating_point_test

evpobr:
Use heap instead of stack.


erikd:
A better solution was commited in
```
commit f224bf9c427be7b1313ab9c3ca98bde79546abad
Date:   Wed Sep 14 17:47:08 2016 +1000

    src/alac.c: Remove large stack allocations
```

PR #174, opened 2016-09-16
CMake: fix external libraries linking

evpobr:
Required libraries are configured properly for `libsndfile`. When you link static `libsndfile` with `add_library`, external libraries and `m` library will be added to target required libraries if needed. Shared `libnsdfile` will not export any libraries, they are not needed in this case.

Application targets are now linked to `m` properly, when `m` is needed and they are linked to shared `libsndfile`, which do not export `m`, `m` will be added directly as its own dependency.


erikd:
Merged. Thanks.

PR #178, opened 2016-09-20
CMake: requre C99 compiler

evpobr:
C99 support is required since we use `inline` keyword and `snprintf` function. The main problem is Visual Studio due to poor C standards support, so versions prior to VS2015 are not dropped.


erikd:
Merged. Thanks.

PR #180, opened 2016-10-05
Cmake clipping test

evpobr:
This PR adds clipping test, just like Autotools toolchain's one.

If cross-compiling, test will not give correct results, so `CPU_CLIPS_POSITIVE` and `CPU_CLIPS_NEGATIVE` values are set to `0` in this case and you can change them in CMake GUI.

`DISABLE_CPU_CLIP` option can be used to disable this feature completely.


erikd:
Merged. Thanks.

PR #184, opened 2016-10-09
Cmake tests part 2

evpobr:
Bunch of tests was added, see individual commits.

CMake's [add_test()](https://cmake.org/cmake/help/v3.0/command/add_test.html) command allows only one command per test, so there are multiple command invocations one by one. I guess it is possible to group them, but i didn't find a proper way to do this. It is not critical, tests works for now.

`win32_ordinal_test` is not implemented yet, it has hardcoded paths to `libsndfile-1.def`, but CMake can be invoked from any location, so this test will fail. I think we need to add optional path parameter to `win32_ordinal_test`.

I'm not sure what is purpose of `pedantic-header-test.sh`, as i think my `sf_version_test` does the same?

May be something other is missing, if you find something i will add it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -746,6 +746,13 @@ if (BUILD_SHARED_LIBS AND LIBM_REQUIRED)
 	target_link_libraries(compression_size_test PRIVATE ${M_LIBRARY})
 endif ()
 
+set (ogg_test_SOURCES tests/ogg_test.c tests/utils.c)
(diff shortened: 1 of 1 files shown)


erikd:
Merged. Thanks.

This seems pretty close to complete now. I need to squash down the commits on this branch and then have a better look at it. Been crazy busy at work, so may not get to this until next weekend.

PR #185, opened 2016-10-10
Do not require autogen & python in build process #154

### LINKED ISSUE #154: Figure out how to auto-generate files on Windows
@evpobr In order to get the Travis build working on Linux and OSX, I had to revert a commit that added auto-generated files to repo. The revert commit was:

```
commit 260af75a1dd47d29eb831335757ae03f2c5bb6cd
Date:   Sun Sep 11 10:15:21 2016 +1000

    Revert "Added in autogened files"
```

The Travis build script now uses `./autogen.sh && ./configure` to generate these files before doing a CMake based build followed by an Autotool based build.

Interested in any suggestions you may have.

evpobr:
PR #185 is how i want to fix #154.

autogen and python are not critical dependencies now.

The main idea of CMake is portability, not only limited to GNU build system and Linux. So we must avoid any external tools and especially Linux specific tools.

autogen is hard stopper, and now it is removed completely to avoid portability problems (yes, Windows). There is no official Windows port, and it is really hard to build it from sources. To keep generated sources in distribution tarballs is half-decision, i didn't see any other projects, that works in this way, everything must just work. May be you don't have PR's from developers under Windows just because they cannot configure propertly from Git. I had many troubles here too, trying to work with LibSndFile under Visual Studio.

python from the other side, is not dropped. It still can be used as optional dependency, to run some tests for example, but it is not used to generate symbol files. Python is really widely used in Linux world, but it is not so in Windows world. I can do this using CMake only, but it will just make life harder. So it is better just to keep symbol files.

So, if we want real portability with CMake, we must do it. Or we will get just another Autotools clone.


bagong:
Would be lovely to get libsndfile building on Visual Studio!

evpobr:
Obsolete, closed.

PR #186, opened 2016-10-11
examples/sndfile-to-text: added '--full-precision' flag

umlaeute:
specifying this flag will store the samples in full precision
(so they can be losslessly restored from the text-file)


erikd:
Thanks @umlaeute !

PR #189, opened 2016-11-13
OS X Build problem MacOS 10.12.1

Pandora-jshimpf:
Running brew as suggested got all the parts but libtoolize was added as glibtoolize.  The autoconf.sh script has code to check for this but it didn't seem to work.  I had to set LIBTOOLIZE=glibtoolize explicitly and then run autoconf.sh to build the configure file.  It ran (not completely see below), the make and make check both worked just fine.

At the end of configure I got.

Installing git pre-commit hook for this project.
./autogen.sh: line 178: .git/hooks/pre-commit: No such file or directory
chmod: .git/hooks/pre-commit: No such file or directory

But as stated above the build went OK.

--jim schimpf


erikd:
@Pandora-jshimpf Why is this a pull request? The commits you reference have nothing to do with the problem you are reporting.

erikd:
The problem you are reporting is fixed in 798edec4e61524977c9e84e on master.

Pandora-jshimpf:
I am an very new to git and I apologize. Thank you for writing.

—jim shimpf

erikd:
@Pandora-jshimpf You probably wanted to open a Github Issue rather than a Pull Request.

I'll close this.

PR #193, opened 2017-01-19
Build system fixes

SoapGentoo:
Hi @erikd 
David from Gentoo again. I've polished up the build system a bit:

- Remove `INSTALL` (part of automake cruft, don't check into VCS)
- `rm -f` doesn't need a check whether the file is there (and this is important, as the `test` command will lead to failure with `make distcheck`)
- Make `_FORTIFY_SOURCE` optional when the compiler already provides it (like with FLAC)
- Make the test-suite generation simpler, more idiomatic and remove running `make` from configure.ac
- Make programs/docs/man pages optional (this is for multilib installations)
- Compile examples only when specifiying `make check`
- Make the test python script py2+py3 compatible. Allowing it to also run on py3 makes it easier on our build systems.

Don't merge it yet, because I believe the test-suite can be simplified further. I think most of the code in the `Makefile.am`s should be factorised out into shell scripts and then run via the `TESTS` variable of Automake. This is the idiomatic and recommended way how to run tests in Automake.

Lastly, what do you think about doing the same FLAC-type approach with CFLAGS? i.e. instead of using a non-standard flag `--disable-gcc-opt`, checking whether `CFLAGS` is set and then based on that. This is also the approach taken by `AC_PROG_CC` internally (http://git.savannah.gnu.org/cgit/autoconf.git/tree/lib/autoconf/c.m4#n508)


evpobr:
@SoapGentoo, for 68facfa there was related pull request #147.

SoapGentoo:
@evpobr I've seen it, but they're not quite identical - yours focuses mainly on getting rid of autogen (which I agree with). My approach at least shifts it to make dist - that still doesnt solve your MSVC problem, but alleviates it slightly.

configure.ac
```diff
@@ -77,7 +77,7 @@ if test x$ac_cv_have_decl_S_IRGRP = xyes ; then
 	AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP],1,[Set to 1 if S_IRGRP is defined.])
 else
 	AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP],0)
-	fi
+fi
```
erikd: Err why? This may not have been your indentation style but its is mine and it is consistent across every shell script in this repo.

programs/test-sndfile-metadata-set.py
```diff
@@ -71,52 +80,52 @@ def make_sine (self):
...
+    print("    %-30s :" % name)
 
 def assert_info (programs, filename, arg, value):
     output = programs.meta_get (False, "%s %s" % (arg, filename))
     if output.find (value) < 0:
-        print "\n\nError : not able to find '%s'." % value
-        print output
+        print("\n\nError : not able to find '%s'." % value)
+        print(output)
         sys.exit (1)
```
erikd: Something about this changes has screwed up the output. Its now:
```
Testing WAV metadata manipulation:
    Empty fail test                :
ok
    Copy test                      :
ok
    Auto date test                 :
ok
    Update test                    :
ok
    Post mod test                  :
ok
    Update test                    :
ok
    Rewrite test                   :
ok
```
The "ok" used to be at the end of the previous line.

SoapGentoo: Fixed

SoapGentoo:
@erikd Amended, the py extraneous line thing should be fixed now, indentation-style restored.

configure.ac
```diff
@@ -507,8 +507,27 @@ if test x$ac_cv_c_compiler_gnu = xyes ; then
...
+					this_is_an_error;
+				#endif
+				}
+			]]
+		)], [
+			AC_MSG_RESULT([yes])
+			MN_ADD_CFLAGS([-D_FORTIFY_SOURCE=2])
+		], [
+			AC_MSG_RESULT([no])
+		])
```
erikd: This should really be in a separate M4 macro file.

configure.ac
```diff
@@ -91,7 +91,7 @@ SHLIB_VERSION_ARG=""
 # Finished checking, handle options.
 
 AC_ARG_ENABLE(experimental,
-	AC_HELP_STRING([--enable-experimental], [enable experimental code]))
+	AS_HELP_STRING([--enable-experimental], [enable experimental code]))
```
erikd: What's the difference between `AC_HELP_STRING` and `AS_HELP_STRING`? Why is the latter better than the former?

Makefile.am
```diff
@@ -8,7 +8,12 @@ if BUILD_OCTAVE_MOD
...
 endif
 
-SUBDIRS = M4 man doc Win32 src $(octave_dir) examples regtest tests programs
+SUBDIRS = M4 Win32 src examples tests
+
+if FULL_SUITE
+SUBDIRS += man doc $(octave_dir) regtest programs
+endif
+
 DIST_SUBDIRS = M4 man doc Win32 src Octave examples regtest tests programs
```
erikd: I think you probably want the `examples` directory inside the `FULL_SUITE` section.

SoapGentoo: I had that initially, but the problem is that when the full suite is disabled, `make check` will fail. Practically, it doesn't matter, as the `check_` prefix prevents anything being built in `examples/`

configure.ac
SoapGentoo: I can't find the direct reference, but `AC_HELP_STRING` is deprecated (and currently an alias to `AS_HELP_STRING`). See for instance https://autotools.io/autoconf/arguments.html
If you google you'll see that X/Gnome have also been switching away.

configure.ac
SoapGentoo: @erikd Submitted as https://github.com/peti/autoconf-archive/pull/107
Once its merged, I'll add it here

configure.ac
erikd: Ok, thanks.

Makefile.am
erikd: I can understand that would be a problem for the `tests` directory, but the `examples` directory contains no tests.

configure.ac
erikd: For now, I'd be willing to have it as a custom macro in the `M4/` directory.

SoapGentoo: I wouldn't worry, Peter Simons merges stuff within a day usually

Makefile.am
SoapGentoo: In https://github.com/erikd/libsndfile/blob/master/programs/test-sndfile-metadata-set.py#L46 the test suite from `programs/` references stuff in `examples/`.
There's two ways going forward:

1. Either leave it as is
2. or put an explicit make dependency in `programs/Makefile.am`, such that `make_sine` is built, whether the full-suite is enabled or not.

Which one do you prefer?

erikd: Ah ok. Lets leave it as is.

SoapGentoo:
@erikd concerning the `--disable-gcc-opt` flag, what do you think about me going the FLAC way?

erikd:
FLAC has many `--disable-xxx` options. Which specific one do you mean? Or more importantly, what are you proposing?

SoapGentoo:
I propose the following:

- No CFLAGS/CPPFLAGS/LDFLAGS provided == `--enable-gcc-opt`
- CFLAGS/CPPFLAGS/LDFLAGS provided by user == `--disable-gcc-opt`

erikd:
I still don't understand exactly what your getting at. For instance I would be very much against dropping all the warning flags that are currently added.

Maybe put up a separate PR just for this CFLAGS change.

erikd:
Thanks @SoapGentoo

PR #196, opened 2017-02-21
get variable number of cues

diemoschwarz:
Hi Erik,

here's a request regarding cue-reading. First, there's a bug in common.c in SF_CUES_VAR_SIZE that leads to memory corruption (shows up in valgrind when there are excactly 100 cues), fixed in 31b6114:

 	fix off-by-one error in SF_CUES size calc 			

Then, the doc mentions reading/writing more than 100 cue markers, but sf_command is too rigid for it, so I'd propose 6ebbe69:

       relax condition to allow query of variable number of cues (psf_get_cues will round down datasize to     num. cues and take minimum of requested and existing cues)

There's also a cue_test.c and some test files for taking or leaving in 3b3c747.
Also, sorry for the bogus .cmake changes, don't know how to get rid of them.

If there's anything I can do to ease checking or applying these patches, please point me to it.
Best...
   ...Diemo


erikd:
This needs to be rebased against current master. There should not be any merge commits.

In fact, since these patches seem to have accumulated over a number of years, it might be easier to make a new branch from my master branch and then cherry pick patches from this one. You should also check that each commit passes the style checker script in `Scripts/cstyle.py` (you can ignore existing errors from this script).

erikd:
You also didn't need to close this PR. You can update the code in a PR by doing a "git force push".

PR #199, opened 2017-03-08
CMake build system

evpobr:
Rebased on current master.


bagong:
Following here (as a SuperCollider builder ;) ) and excited to see this happening! Just built successfully on OSX and with VS 2015/64 (!). Still trivial errors in install, but it's getting close! Will we be able to include Flac and Ogg on Windows easily?
Biiiig thanks!

evpobr:
> Will we be able to include Flac and Ogg on Windows easily?

@bagong , it is possible, but not easy. You must build `libogg`, `libflac` and `libvorbis` before. Then copy headers and libraries like this:

    c:\libogg\include\ogg <- headers
    c:\libogg\lib <- lib file
    c:\libvorbis\include\vorbis <- headers
    c:\libvorbis\lib <- lib file
    c:\libflac\include\flac <- headers
    c:\liblac\lib <- lib file

and so on for `libvorbis` and `libflac`.

Then run CMake GUI and configure `libsndfile` project. You will see that all those libararies are not found. It happens because Windows doesn't have system folders to store headers and libraries. So you must help CMake to find them.

Add `OGG_ROOT`, `VORBIS_ROOT` and `FLAC_ROOT` variables to CMake GUI and set set them like this:

    OGG_ROOT = c:\libogg
    VORBIS_ROOT = c:\libvorbis
    FLAC_ROOT = c:\libflac

Run configure again. CMake will finally find everything and enable Ogg, Vorbis and Flac support.

As i know, at least `libogg` and `libvorbis` have CMake projects. Then you can configure them and install somewhere, required directory structure will be made automatically. But you still need to set up `XXX_ROOT` variables.

My be i missed something, but you can try.

evpobr:
> Just built successfully on OSX and with VS 2015/64 (!).

DLL works under Windows? It requires DEF file to export functions, but i didn't implement its generation yet.

bagong:
@evpobr , whether the result can already be used is a different question ;). I haven't gotten there yet, and can't dedicate time to this at the very moment. But I'll do so in the coming days. Here I just wanted to give strong thumbs up and show that the efforts aren't passing unnoticed. ;) The option to build easily in native Windows using MSVC is much appreciated. Portaudio is already there, and there is some movement on the FFTW3-front ;) So there is a nice trend in the important libre audio libraries that come from a unix background.

erikd:
> create_symbols_file.exe can also create export definitions for Linix, Darwin (MacOS) and OS/2 but CMake doesn't support it. This is not a big problem, compilers for these OSes will export all functions by default.

I think this needs fixing. The autotools build chain does not export all symbols by default on Windows, Mac or Linux. Its almost certainly possible to make CMake use the same symbols files used by the autotool build chain.

evpobr:
I think it is ready now to merge. Will add some non critical features later:

* MacOS installation
* Some options from `configure.ac` - (`ENABLE_OCTAVE`, `ENABLE_BOW_DOCS` and so on)

Some things will not be implemented now:

* Tests for Windows (MSVC) - no, `autogen` is stopper here.
* Restrict functions to export from shared libraries under Linux, MacOS & etc - no, CMake doesn't support it.

evpobr:
> I think this needs fixing. The autotools build chain does not export all symbols by default on Windows, Mac or Linux. Its almost certainly possible to make CMake use the same symbols files used by the autotool build chain.

No, CMake supports natively symbol files under Windows only (DEF files). In theory we can add requred compiler flags directly, but it is not easy and not CMake way. If you really want this i can try.

evpobr:
@erikd - MacOS problems:

```
Undefined symbols for architecture x86_64:
  "// Auto-generated by create_symbols_file", referenced from:
     -exported_symbol[s_list] command line option
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Any thoughts?

Ok, i have idea. You don't break ABI compatibility under Windows, so let's use DEF files under Windows. It is already implemented in `autotools` scripts and trivial to use with CMake. But may be we can use visibility trick under all other OSes? It will break nothing.

How it should be. In `sndfile.h`:

```
#if defined SNDFILE_EXPORTS
  #define SNDFILEAPI __attribute__ ((visibility ("default")))
#endif
```

And for any ABI function:

`SNDFILEAPI SNDFILE* 	sf_open		(const char *path, int mode, SF_INFO *sfinfo) ;`

Then we need to add `-fvisibility=hidden` to compiler. That's all. We can even still use symbol files in `autotool` chain. And this feature is supported by CMake natively.

erikd:
> Ok, i have idea. You don't break ABI compatibility under Windows, so let's use DEF files under Windows. It is already implemented in autotools scripts and trivial to use with CMake. But may be we can use visibility trick under all other OSes? It will break nothing.
>
> How it should be. In sndfile.h:
>
> #if defined SNDFILE_EXPORTS
>  #define SNDFILEAPI __attribute__ ((visibility ("default")))
> #endif

It should be possible to put it in `common.h` which IMO is neater.

Lets try this on a separate branch off master first, just with the autotools build system.

evpobr:
@erikd, i have not enough skills to implement this with autotools. Can you write required code? We must check if system is not Win32 and then check if compiler supports `fvisibility=hidden` option, then define `SNDFILE_EXPORTS` for shared library build.

erikd:
OK, i'll give that a go some time this weekend.

evpobr:
@erikd, i found useful macro, trying to use it.

evpobr:
PR #200 , you can do better.

madebr:
Hey, you use `${CMAKE_CURRENT_SOURCE_DIR}` as target files in `configure_file` and `add_custom_command`.
Maybe use `${CMAKE_CURRENT_BINARY_DIR}`?
That way out of tree builds become possible.

My fork https://github.com/madebr/libsndfile/tree/cmake supports out of tree builds.

evpobr:
There is really obsolete CMakeLists.txt at master. Look at PR #199.

bagong:
Some of the details discussed here I don't understand, I will need someone to take my hand if more thorough testing by an outsider is wished. What I've done so far: build libsndfile (without subsidiary libraries) successfully with VS2015, used the lib and dll for dynamically linking to SuperCollider and done a 4-channel wav-recording there that seems to be written nicely. I am happy to do more if you tell me what would be helpful. For now I'll try to get Flac and sisters into the game.

evpobr:
@bagong , that thing you didn't understand is main stopper, then i hope @erikd will merge my pull request to his master branch. And then we will polish other things. Now just inspect my `CMakeLists.txt` code, do pull requests. That one about missing docs was really useful.

bagong:
It works: ogg/flac/vorbis build successfully out of the box with VS2015, can be linked in statically and I ended up with a libsndfile-1.dll (as well as a yet untried static library) that behaves in SuperCollider like the distributed version, i.e. we can read and write uncompressed and read ogg and flac. That's a nice base for enhancing things in SC.
I observed that there are subtle differences building with cmake and directly from VS in the external libs where that is possible. When building Ogg with cmake and trying to link it into flac (which needs to be built with VS) I got a single symbol not found error in FLAC. If everything directly in VS no problems like this.

evpobr:
New visibility control implemented via 9dff3e5. Public header is untouched, `SNDFILE_API` is defined in private header `src/common.h`. DEF file is still in use under Win32, ABI is not changed. Autotools build chain is not modified.

@erikd, please check exported symbols of CMake's shared library.

src/ALAC/ag_dec.c
```diff
@@ -26,6 +26,10 @@
 	Copyright:	(c) 2001-2011 Apple, Inc.
 */
 
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
```
erikd: Why is this needed?

evpobr: Commit 267285f.

    #if HAVE_CONFIG_H
    ...
    #endif

Old stuff. Probably from times, when i thought not to configure `config.h` and pass required defines directly to compiler with `/D` switch.

`#include "config.h` is required here because some old compilers doesn't support `__inline` keyword.  It is VS2013 and before, some old GCC versions.

I guess it can be removed. VS2013 has bad C support anyway. VS2015 is much better and will work without it.

evpobr:
@erikd I can make another PR squashing some commits by functionality and removing old stuff.

evpobr:
@erikd, `.travis.yml` can be simplified, some `trusty` packages were upgraded. CMake is now 3.2.2, GCC 4.8.4, Clang - 3.5.

erikd:
If you do that, please do it on PR #205 within the next 24 hours (I won't have sufficient time to look at it before then).

evpobr:
I will not do it now, I'm waiting for you to finish the code review.

erikd:
> I will not do it now, I'm waiting for you to finish the code review.

I'm not going to start the code review until I know I have sufficient time to get through it in one sitting. That will not be until the weekend. Now is a perfect time to do a minor update like the one to `.travis.yml`.

PR #200, opened 2017-03-11
Add API exported functions visibility control

evpobr:


erikd:
It looks like the `SNDFILEAPI` stuff doesn't need to be in the public header file. Its working for me in Linux withe all that stuff confined to `src/sndfile.c`. Need to do some more testng.

evpobr:
It will work under Linux probably. Under Windows (MSVC) `SNDFILEAPI` should  `__declspec(dllimport)` when using header. so it will not work.

evpobr:
Of course It is ok if we use DEF for Windows.

erikd:
Commit 9dff3e5 seems to be a version of this patch that doesn't change to public header file. Would you be able to update this PR to that version of the patch?

evpobr:
This outdated PR is just proof of concept how it works with Autotools. Do you want this feature in Autotools build chain? If so, i will update it.

The real work is done in #199 with 9dff3e5 to make CMake toolcain work properly. That commit was referenced here because it is related.

erikd:
Yes, please. Small self-contained PRs like this are much easier to review and validate and can go straight onto master.

evpobr:
That's why i think #199 should be merged to master, as experimental feature.

erikd:
#199 Will require a *huge* amount of work before its ready for master. This is is ready now.

evpobr:
Did you mean it is not ready yet?

erikd:
It mostly means there's lots of work for me to do.

It means there a dozens of trivial commits that should be squashed into a single commit.

It means I need to look at every change to every file to make sure I think its correct.

bagong:
@evpobr , using `git rebase -i`could help to make this PR more digestible for Erik.

evpobr:
Changes were accepted to master with 1bff04b, CMake related code was added to #205.

bagong:
Exciting dynamic! ;)

PR #205, opened 2017-03-14
Rework CMake build system

evpobr:
Squashed version of #199.

PR #210, opened 2017-03-24
CMake: remove dependency on autotools

madebr:
Hey,
this pull request adds the following features:
- autogen, autoheader, grep, sed and wc are replaced with CMake implementations
- out-of-tree builds: generated sources and header files are put in the build directory. The source tree remains pristine.
- autogen is replaced with a self-written cmake implementation: See https://github.com/madebr/CMakeAutoGen for more details


erikd:
There is already a much more significant rework of the CMake stuff in #205.

madebr:
Ok, I'll rebase this PR onto that patch set once it hits master.
I think some commits of this PR add value.

madebr:
The `cmake-config-revised`  branch does not exist (yet).
Should I wait until #205 lands?

evpobr:
> The `cmake-config-revised` branch does not exist (yet).
> Should I wait until #205 lands?

@madebr , To avoid branching hell it's better to wait, #205 is ready to merge, i don't push new commits there, but i can make force push.

@erikd , is there an approximate release date for v1.0.28?

erikd:
I'd say with the next 7 days.

PR #218, opened 2017-04-05
configure.ac: Remove useless `LIBTOOL_DEPS` macro

evpobr:
It is not required when using libtool with automake, as we do.

Proof: http://tinf2.vub.ac.be/~dvermeir/manual/autobook/autobook-1.2/autobook_87.html


erikd:
Merged. Thanks.

PR #221, opened 2017-04-06
CMake: bump version up to 1.0.29pre1

evpobr:
Issue: https://github.com/erikd/libsndfile/issues/71

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required (VERSION 3.1.3)
 
-project(libsndfile VERSION 1.0.28)
+project(libsndfile VERSION 1.0.29)
 
 #
 # Variables


erikd-ambiata:
Is CMake not able to handle a version like "1.0.29pre1" ?

Ideally I'd like to have the version number in one place and have both build systems use it.

erikd:
Is CMake not able to handle a version like "1.0.29pre1" ?

Ideally I'd like to have the version number in one place and have both build systems use it.

evpobr:
https://cmake.org/cmake/help/v3.0/command/project.html:

> If VERSION is specified, given components must be non-negative integers.

erikd:
Maybe it should then be `1.0.29.1`

evpobr:
I think `1.0.29.1` means the first bugfix release of `1.0.29` official release, not the first development version.

erikd:
And `1.0.29` means an official release. That means it would be more correct for the CMake version number to be the last release number rather than the next release number.

evpobr:
> And 1.0.29 means an official release. That means it would be more correct for the CMake version number to be the last release number rather than the next release number.

No. CMake is not Autotools.

Your current libsndfile version in `configure.ac` is `1.0.29pre1`. It means version `1.0.29` in `pre1` stage, not `1.0.29.1`.

CMake version is the same. Now it is 1.0.29. It means it can be release, `beta2`, `pre1`, we don't know, but it is 1.0.29, CMake's naming scheme is absolutely enough to perform version checks and that's the only purpose of `PROJECT_VERSION`.

From the other side, sometimes we want to use stage postfix, to show exactly what it is. And then we introduce `CPACK_PACKAGE_VERSION` variable with `1.0.29pre1`. It goes to config.h, it used to construct distribution zip filename (`libsndfile-1.0.29pre1-win64.zip`) etc.

I don't know is it feature or bug of CMake. But that's how it works.

Also i have plans to implement version sharing between Autotools and CMake (parse `AC_INIT` of `configure.ac`). But i want to fix discovered bugs first.

evpobr:
@erikd , could you allow me to commit & merge changes related to CMake directly to master? I see you're very busy. It's too long to wait for a week to merge such a trivial commit.

erikd:
I'm still not happy with this inconsistency between the way autotools and Cmake handles version numbers.

erikd:
Maybe the Cmake version could be `1.0.29.1` with comment above saying this is equvialent to `1.0.29pre1`.

evpobr:
Did you read my anwser 3 days ago? It is not equvialent. I'm absolutely sure this commit should be exactly as it is. Just believe me as CMake developer of your project. Your proposal will break the versioning.

erikd:
With your commit CMake builds something labeled `1.0.29` which is *not* version `1.0.29`. Cmake is wrong. We need to find some solution to this that does not involve giving the build artifact the incorrect version number.

evpobr:
> We need to find some solution to this that does not involve giving the build artifact the incorrect version number.

Please, read again my answer 3 days ago.

Making CMake distro ZIP with CPack produces `libsndfile-1.0.29pre1-win64.zip`. You can try.

erikd:
This is not about the tarball. This is about the library version numbers.

If I apply you PR to a branch and run `Scripts/cmake-build.sh`, the library binary that results is `libsndfile.so.1.0.29`.

evpobr:
That makes sense, thanks.

evpobr:
I've got news for you: Autotools result is also `libsndfile.so.1.0.29` 😄 

Just tested under [WSL](https://msdn.microsoft.com/en-us/commandline/wsl/about) (based on Ubuntu Trusty).

Autotools build:

autotools-v1 0 29pre1

CMake build:

cmake-v1 0 29pre1

I'm confused. Can you confirm it?

erikd:
Wow, I was pretty sure this used to keep the `preX` on the shared lib versions. This must have changed in a recent version of the autotool without me noticing.

At least the version number in the binary is correct:
```> strings _Build/libsndfile.so.1.0.29 | grep ^libsndfile-1.
libsndfile-1.0.29pre1
```
I'll merge this.

erikd:
Merged. Thanks.

PR #226, opened 2017-04-10
CMake: remove dependency on autotools

madebr:
Hey,
this pull request adds the following features:
- out-of-tree builds: generated sources and header files are put in the build directory. The source tree remains pristine. This allows parallel builds of libsndfile binaries with different configurations. (`src/config.h` and `src/sndfile.h` might change when using different CMake switches)
- autogen is replaced with a self-written CMake implementation: See https://github.com/madebr/CMakeAutoGen for more details

- I created a static library containing only `tests/utils.c` because using the output of `add_custom_command` in more than one independent target causes build conflicts. See [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html).

I can reliably reproduce this bug by executing the following commands in current master:
```
cd $LIBSNDFILEDIR
rm tests/utils.[ch]
cd Build
rm -r *
cmake ..
make -j20
```

The parallel compilation will fail, although `make` succeeds.


evpobr:
> I created a static library containing only tests/utils.c because using the output of add_custom_command in more than one independent target causes build conflicts. See add_custom_command.

We need to use `add_custom_target`. Can you implement it?

evpobr:
@madebr , great work, it i will check it tomorrow. Leave this PR as it is. If you have time please check `add_custom_target` approach in another PR. Interestring how it should be. We will choose the best.

@erikd , what do you think about 357115b?

evpobr:
Tested under Windows with VS2017:

* @madebr , Tests framework is not built and files are not generated if autogen is missing. Remove [these](https://github.com/madebr/libsndfile/blob/master/cmake/SndFileChecks.cmake#L194-L198) lines in `cmake/SndFileCheck.cmake`
* `pcm_test.c` is not generated, no `pcm_test` target to trigger generation. It is my mistake. @madebr , can you add that test to your PR?
* Discovered problem if Autootols build was first, generating `config.h` and `sndfile.h` in `src/`. If we run CMake then, we have two copy of those files, and complier uses headers from `src/`. Don't know is it bug. Maybe put warning in README to `make clean` before using CMake.
* All other autogenerated files are identical to autogen's ones. Bit to bit.

@erikd: there are many build erros in tests to be fixed later:

* Well known 'missing unistd.h';
* undeclared [`M_PI`](http://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio)
* undeclared `off_t` in `test_main.c` - `#include <sys/types.h>` helps
* undeclared [`STDIN_FILENO`](http://stackoverflow.com/questions/13531677/stdin-fileno-undeclared-in-windows)

evpobr:
Thanks @madebr.

PR #239, opened 2017-04-16
CMake polishing

evpobr:
# Features

* Autogenerating sources without AutoGen! Great thanks to @madebr
* True out-of-tree builds (thanks @madebr). All generated builds are placed to separate build directory
* Add `pcm_test` and `win32_ordinal_test` to test suite
* Add examples and regtest targets
* Testing improvements - CMake internal test support files are not generated if you disable testing
* Configurable [CPack](https://cmake.org/Wiki/CMake:Packaging_With_CPack) support - you can disable it now with `ENABLE_CPACK` option set to `OFF`
* Add `ENABLE_BOW_DOCS` option. Set to `ON` to get black on white documentation style
* `ENABLE_STATIC_RUNTIME` option - set to `ON` to enable static runtime with MinGW and [Visual Studio](https://msdn.microsoft.com/en-US/library/2kzt1wy3.aspx).
* Add [config package support](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#package-layout). Now you don't need to write your own FindLibSndFilecmake find module. Usage is as simple as:
```cmake
find_package (LibSndFile 1)
if (LIBSNDFILE_FOUND)
    taget_link_library (<your_target> PRIVATE sndfile-static)
endif ()
```
This work under Windows with default install location `%PROGRAMFILES%`. If you want to install LibSndFile to different location, use `LibSndFIle_DIR` option to set proper path.

# Not to be implemented

* `generate` target. Many quirks are required.
* Octave module. Really complicated detection procedure.
* [Package registry](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#package-registry). Package config mode is moslty the same. Install to arbitrary location and use `LibSndFIle_DIR`.

# Screenshots

cmake

vs2017

@erikd - tell what to rebase\squash please.

PR #242, opened 2017-04-24
Fix CMAKE_CURRENT_SOURCE_DIR

Iunusov:
CMAKE_SOURCE_DIR can return wrong root, when library is used as dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ if ((NOT BUILD_STATIC_LIBS) AND (NOT BUILD_SHARED_LIBS))
 	set (BUILD_STATIC_LIBS ON)
 endif ()
 
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/")
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/")
 
 #
 # Setup definitions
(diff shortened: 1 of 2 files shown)


evpobr:
@Iunusov, i hope that was fixed in 15fa0f4 of #239.

Iunusov:
Yes, but I still getting error with your commit (cmake/TestInline.cmake is
not changed at all, but it also uses CMAKE_SOURCE_DIR instead of
CMAKE_CURRENT_SOURCE_DIR). Let's check your CmakeLists.txt and
cmake/TestInline.cmake.

evpobr:
Ok, `TestInline.cmake` fixed, try now. It was forced push, reset last 3 commits and pull again. I know it is bad, but `wip/cmake` is almost ready, i don't want to add new commits.

Another question: do you use `add_subdirectory` to add LibSndFile project?

Iunusov:
yes, I am using add_subdirectory() and Visual Studio 2015.

I will check.

Iunusov:
checked, still getting errors because of wrong root:

-- Checking for inline...
*CMake Error: File D:/WIndows/Desktop/test/cmake/TestInline.c.in
<http://TestInline.c.in> does not exist*.
CMake Error at deps/libsndfile/cmake/TestInline.cmake:5 (configure_file):
  configure_file Problem configuring file

Real libsndfile location: D:\WIndows\Desktop\test\*deps*\libsndfile

evpobr:
I'll check it.

Iunusov:
Thanks.

evpobr:
Yes, there are many issues when adding LibSndFile with `add_directory`. Most of them are fixed in `bugfix/cmake-out-of-source`. But tests are not working, disable them with `BUILD_TESTS=OFF`. Now i need @madebr help to fix AutoGen implementation and tests. It will take time.

Another way is to build and install `libsndfile` and use `find_package(LibSndFile)`. It works with `master` branch.

evpobr:
I made PR #245, hope it fixes all problems.

Iunusov:
I will check, thanks!

Iunusov:
Checked, now it works fine without the issues.

evpobr:
Ok. Changes are merged to master barch.

PR #246, opened 2017-04-26
Autotools: Implement symbols visibility control

evpobr:
* Use symbol file under Win32 with MinGW only
* Use unified visibility control for other platforms
* Add recommended `win32-dll` option to `LT_INIT` macro


src/libsndfile-1.def
```diff
@@ -1,6 +1,4 @@
-; Auto-generated by cmake
-
-LIBRARY ${PROJECT_NAME}-${PROJECT_VERSION_MAJOR}.dll
+LIBRARY libsndfile-1.dll
```
erikd: Is `LIBRARY ${PROJECT_NAME}-${PROJECT_VERSION_MAJOR}.dll` not correct? I think it is. I don't see any reason to make this change (or the similar one above).

evpobr: It was correct.

I thought that this file is changing very rarely, so there is not much point in configuring it every time.

But of course i can make a single template for an Autotools and CMake and configure it with something like `PACKAGE` and `VERSION_MAJOR`.

I know how to do this with CMake, but can you give me `VERSION_MAJOR` regex? I see `VERSION_MINOR` in `configure.ac`, something like that. I'm not very friendly with regular expressions.

erikd:
```
VERSION_MAJOR=`echo $PACKAGE_VERSION | $SED "s/\..*//"`
```

evpobr: @erikd:

        ; Auto-generated by @GEN_TOOL@

`GEN_TOOL` for CMake is set to `cmake`. `autoconf` for Autotools is OK?

erikd: Sounds reasonable.

erikd:
Should we wait for fixed Appveyor builds for this?

evpobr:
Small fix just to remove AppVeyor test results. I guess we can merge it right now.

PR #250, opened 2017-04-28
CMake: Fix missing libsndfile-1.def

evpobr:


bagong:
Confirmed to fix the issue.

PR #252, opened 2017-04-29
CMake: Win32 fixes

evpobr:
* Fix linking with shared library (DEF file was not used)
* Version script was not used
* Use the same version script for Autotools and CMake
* Give proper name to shared library import lib
* Public headers were not installed on static library only build

Closes: https://github.com/erikd/libsndfile/issues/249


erikd:
If raise a PR, there is no need to request a review. That will result in one less email in my inbox.

PR #261, opened 2017-05-08
Fix sndfile-regtest compilation with MSVC

evpobr:
`getcwd` function requires `direct.h` header.

### DIFF
--- a/regtest/database.c
+++ b/regtest/database.c
@@ -27,6 +27,9 @@
 #endif
 #include <string.h>
 #include <fcntl.h>
+#ifdef HAVE_DIRECT_H
+#include <direct.h>
+#endif
 #include <sys/stat.h>
 
 #include <sndfile.h>


gvanem:
The `sndfile-regtest` is crap! It does not really support Windows since it wants all paths to have only `/` slashes in them.

E.g. `bin\sndfile-regtest.exe --add-file \gv\dx-radio\Coders-Decoders\codec2\wav\all.wav` 👍 

```
sndfile-regtest : using libsndfile-1.0.29 (10-September-2020)

    \gv\dx-radio\Coders-Decoders\codec2\wav\all.wav : could not open : Not a valid SNDFILE* pointer.
File : F:\MingW32\src\Sound\libsndfile/\gv\dx-radio\Coders-Decoders\codec2\wav\all.wav
```

I'm working on a patch.

evpobr:
I don't know if `sndfile-regtest` is needed at all.

PR #269, opened 2017-05-20
rework gitignore to cover msvc builds

mancoast:


evpobr:
Thanks, we can take this as basic VS ignore rules, But there are much more to [ignore rules](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore).` Also CMakeSettings.json`, stuff from VS2017.

I use VS2017 Comminuty Edition in Open Folder mode. It supports CMake and hides all the generated files, so Git tree is clean.

PR #274, opened 2017-05-25
Fix broken ABI

evpobr:
Unfortunately ABI was broken when symbols files were removed
in favour of new visibility control of exported functions.

Visibility control with -fvisibility works fine, but symbol
scripts had another feature - versioned symbols. And we lost it.

Since we can not make our symbols to be versioned  with the new
approach, it's decided to return everything back.

* CMake: Restore symbol files generation
* CMake: Python is required to build shared libraries
* Autotools: Restore symbol files generation


erikd:
Thanks @evpobr

PR #282, opened 2017-06-04
Autotools cleanup

SoapGentoo:
* Modernise the Autoconf code using `AS_*` macros
* Rely more on `AX_*` macros instead non-standard `MN_*` macros


evpobr:
@SoapGentoo , looks nice. Not sure about Makefile.am in `M4/` dir, can you explain?

SoapGentoo:
@evpobr the M4 stuff is purely for Autoconf, there's no reason `make` needs to go into that dir.

SoapGentoo:
@erikd this also implements the flag handling as discussed previously.

So say a user just runs vanilla `./configure`. He then ends up with

```
CFLAGS : ..............................  -Wall -Wextra -Wpointer-arith -Wcast-align -Wcast-qual -Wshadow -Wwrite-strings -Wundef -Wuninitialized -Winit-self -Wdeclaration-after-statement -Wvla -Wbad-function-cast -Wnested-externs -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations -Waggregate-return -O2 -pipe 
CXXFLAGS : ............................  -Wall -Wextra -Wpointer-arith -Wcast-align -Wcast-qual -Wshadow -Wwrite-strings -Wundef -Wuninitialized -Winit-self -Wctor-dtor-privacy -Wnon-virtual-dtor -Woverloaded-virtual -Wreorder -Wsign-promo -O2 -pipe
```
Notice how the `-O2 -pipe` was added. No power users add their own flags with `./configure CFLAGS="-Ofast" CXXFLAGS="-Ofast"`

```
CFLAGS : .............................. -Ofast -Wall -Wextra -Wpointer-arith -Wcast-align -Wcast-qual -Wshadow -Wwrite-strings -Wundef -Wuninitialized -Winit-self -Wdeclaration-after-statement -Wvla -Wbad-function-cast -Wnested-externs -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations -Waggregate-return 
CXXFLAGS : ............................ -Ofast -Wall -Wextra -Wpointer-arith -Wcast-align -Wcast-qual -Wshadow -Wwrite-strings -Wundef -Wuninitialized -Winit-self -Wctor-dtor-privacy -Wnon-virtual-dtor -Woverloaded-virtual -Wreorder -Wsign-promo
```

Notice how `-O2 -pipe` was not added to `CFLAGS` and `CXXFLAGS`.

src/sfconfig.h
```diff
@@ -113,9 +113,6 @@
 #define HAVE_X86INTRIN_H 0
 #endif
 
-#define CPU_IS_X86		(defined __i486__ || defined __i586__ || defined __i686__ || defined __x86_64__)
-#define CPU_IS_X86_64	(defined __x86_64__)
-
```
erikd: I'd prefer:
```
#if (defined (__i486__) || defined (__i586__) || defined (__i686__) || defined (__x86_64__))
#define CPU_IS_X86		  1
#define CPU_IS_X86_64	  0
#elif defined __x86_64__
#define CPU_IS_X86		  0
#define CPU_IS_X86_64	  1
#else
#define CPU_IS_X86		  0
#define CPU_IS_X86_64	  0
#endif
```
and the rest of the code to stay the same. This should probably be in its own PR.

configure.ac
```diff
@@ -76,11 +76,11 @@ AC_CHECK_HEADERS(sys/time.h)
...
-if test x$ac_cv_have_decl_S_IRGRP = xyes ; then
-	AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP],1,[Set to 1 if S_IRGRP is defined.])
-else
-	AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP],0)
-	fi
+AS_IF([test "x$ac_cv_have_decl_S_IRGRP" = "xyes"], [
+	AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP], [1], [Set to 1 if S_IRGRP is defined.])
+], [
+	AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP], [0], [Set to 0 if S_IRGRP is not defined.])
+])
```
erikd: I think that idiom is far less clear that the original and find the indentation you have chosen to be less than ideal.

Since this is a common idiom in libsndfile's `configure.ac` I would prefer if we defined a new M4 macro that expands to what you have with improved indenting. Eg:
```
AS_IF([test "x$ac_cv_have_decl_S_IRGRP" = "xyes"], [
    AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP], [1], [Set to 1 if S_IRGRP is defined.])
    ], [
    AC_DEFINE_UNQUOTED([HAVE_DECL_S_IRGRP], [0], [Set to 0 if S_IRGRP is not defined.])
    ])
```

configure.ac
```diff
@@ -53,6 +53,8 @@ LT_PROG_RC
 AC_PROG_INSTALL
 AC_PROG_LN_S
 
+AM_PATH_PYTHON
+
```
erikd: This could easily be a separate PR and would make reviewing of the whole easier.

Makefile.am
```diff
@@ -40,15 +40,15 @@ test: check-recursive
 
 # Target to make autogenerated files.
 checkprograms :
-	(cd src ; $(MAKE) libsndfile.la checkprograms)
-	(cd tests ; $(MAKE) checkprograms)
+	$(MAKE) -C src libsndfile.la checkprograms
+	$(MAKE) -C tests checkprograms
```
erikd: I am aware of this way of doing it, but why is this better? What happens if I accept this patch and later get a patch to change it back?

configure.ac
```diff
@@ -28,7 +28,7 @@ AC_CANONICAL_TARGET([])
 
 AC_CONFIG_HEADERS([src/config.h])
 
-AM_INIT_AUTOMAKE
+AM_INIT_AUTOMAKE([foreign dist-bzip2 no-dist-gzip subdir-objects])
```
erikd: Happy to have this as a separate PR as well.

configure.ac
```diff
@@ -473,61 +483,28 @@ AS_IF([test "x$enable_bow_docs" = "xyes"], [
...
+	XIPH_GCC_STACK_PROTECTOR
+	XIPH_GXX_STACK_PROTECTOR
+])
 
-	# -Winline -Wconversion "
-	CFLAGS="$CFLAGS     $common_flags -Wbad-function-cast -Wnested-externs -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations -Waggregate-return"
-	CXXFLAGS="$CXXFLAGS $common_flags -Wctor-dtor-privacy -Wnon-virtual-dtor -Woverloaded-virtual -Wreorder -Wsign-promo"
+AS_IF([test "x$enable_test_coverage" = "xyes"], [
+	AX_APPEND_COMPILE_FLAGS([-coverage], [CFLAGS])
+])
```
erikd: There is an existing indentation scheme in this file and in this project that should be adhered to. Eg:
```
AS_IF([test "x$enable_test_coverage" = "xyes"], [
	AX_APPEND_COMPILE_FLAGS([-coverage], [CFLAGS])
	])
```

Makefile.am
SoapGentoo: 1. you don't check whether cd fails
2. the (...) construct opens a subshell, which should generally be avoided
3. `make -C` fulfills both of the previosu requirements

PR #296, opened 2017-06-19
Fixes v3

SoapGentoo:


SoapGentoo:
@erikd Fixed now, all tests are run in parallel now in Automake, which does give it a bit of a boost.

erikd:
I'm fine with everything being built in parallel, but not happy with the way the test output has changed. 

* There are tests in the `src/` directory which are not being run (and maybe not even being built).
* The tests used to terminate on the first error which is important for the debugging processes.
* The test output is hidden.

The test output was the way it was because that made development, testing and debugging easier.

SoapGentoo:
1. I actually checked that all tests are being built and those that used to be run still run. Some of the `check_PROGRAMS` weren't added to `TESTS`, which I have added now.
2. I get that terminating on first error is useful, but given the performance improvement, reporting _all_ errors seems just as useful?
3. That's a feature, not a bug. There's no way to run tests in parallel and provide non-intermingled stdout at the same time without complicated double-buffering, which I take is way too complex for a make-based build system. You can always check for errors in the per-test log files?

Anyhow, I've reenabled the serial test harness, such that tests are run serially and their output send to stdout.

src/Makefile.am
```diff
@@ -1,132 +0,0 @@
...
-		ALAC/ag_enc.c ALAC/dp_dec.c ALAC/dp_enc.c ALAC/matrix_dec.c \
-		ALAC/matrix_enc.c ALAC/alac_decoder.c ALAC/alac_encoder.c
-
-#===============================================================================
-# Test programs.
-
-test_main_SOURCES = test_main.c test_main.h test_conversions.c test_float.c test_endswap.c \
-					test_audio_detect.c test_log_printf.c test_file_io.c test_ima_oki_adpcm.c \
-					test_strncpy_crlf.c test_broadcast_var.c test_cart_var.c \
-					test_binheader_writef.c
```
erikd: This test program still seems to be missing from the non-recursive build. This test is in the `src/` directory because it requires access to symbols that are not public in the shared object / DLL.

src/Makefile.am
```diff
@@ -1,132 +0,0 @@
...
-test_main_LDADD = libcommon.la
-
-G72x_g72x_test_SOURCES = G72x/g72x_test.c
-G72x_g72x_test_LDADD = G72x/libg72x.la
-
```
erikd: This is another.

src/Makefile.am
SoapGentoo: Add here https://github.com/erikd/libsndfile/pull/296/files#diff-c949f93d03f44a4217d7a138f9e2e54aR86

src/Makefile.am
SoapGentoo: Add here https://github.com/erikd/libsndfile/pull/296/files#diff-c949f93d03f44a4217d7a138f9e2e54aR113

erikd:
Still not keen on the changes to the test output. I am more than fine with the non-recursive automake stuff but the test output changes add nothing and some of the tests (like `regtest/sndfile-regtest` and `tests/win32_test` were never intended to be part of the test suite regardless of what their file names may suggest.

SoapGentoo:
@erikd I've restored the old test-wrapper script with some of my minimal fixes.

erikd:
Sorry to be a pain in the neck, but there's still things broken with this patch. An important part of the existing build system is cross-compiling to Windows and running the test suite under Wine.

If I run:
```
Scripts/linux-to-win-cross-configure.sh w64
make clean all check
```
I get:
```
  CC       programs/sndfile-play.o
  CCLD     programs/sndfile-play.exe
/usr/bin/x86_64-w64-mingw32-ld: cannot find -lasound
collect2: error: ld returned 1 exit status
```
which is obviously not correct. It should not be linking libasound when cross compiling to Windows.

There are a bunch of other odd configurations I build every now and then (like native builds for OpenBSD, FreeBSD and cross-compiling to Android). I'd really prefer it if none of those break.

SoapGentoo:
@erikd ok, finally found some time to fix it.

1. In everything that I do, I never intend to break cross-compilation/portability to other OSes. The failure comes from `pkg-config` falling back to the system ALSA and not the Windows/MinGW one (which doesn't and cannot exist). In detecting ALSA, I've added the compile check back after inferring the parameters via `pkg-config` which should error out internally and clear the `ALSA_CFLAGS` and `ALSA_LIBS` variables. Please give it a try.
2. I've slightly adjusted your GNU prefix rules, in order for the `.h` targets to invoke the `.c` targets directly (this is better, as `make` then knows it recurses into another rule and prevents race conditions, see also https://www.gnu.org/software/automake/manual/html_node/Multiple-Outputs.html).

erikd:
There is still a problem wit the command:
```
Scripts/linux-to-win-cross-configure.sh w64
```
which terminates with:
```
config.status: error: cannot find input file: `Scripts/build-test-tarball.mk.in'
```

SoapGentoo:
@erikd restored the file (I thought it wasn't necessary anymore with `make distcheck`)

erikd:
Sorry this took so long.

SoapGentoo:
@erikd Thanks, was my fault too :smile: Ping me if stuff broke.

erikd:
@SoapGentoo If for instance I change the file `src/flac.c` how do I build the library and just one of the tests? For instance, I used to be able to do this:
```
(cd src && make) && \
(cd tests && make header_test) && \
tests/header_test flac
```
for super fast turn around times when debugging.

PR #306, opened 2017-08-03
Fix CPU detection regression

madebr:
On Intel cpus, CPU_IS_X86_64 was always defined to 0.

* regression introduced in 8321896954d6e53e9b2bdf025ce8e7202c0fd7d3


erikd:
You are right, that the code is not correct, but unfortunately your fix is not correct either, not is the code *before* commit 8321896 .

I believe the version on https://github.com/erikd/libsndfile/pull/308 is what I originally intended.

PR #307, opened 2017-08-07
Make sure FLAC__stream_encoder_init_stream() is only called once.

hzeller:
The FLAC__stream_encoder_init_stream() function complains if it is
called multiple times. If we happen to have flushed the header
manually before with
 sf_command(snd, SFC_UPDATE_HEADER_NOW, ...)
.. then all subsequent sf_write...() functions fail otherwise because
they attempt to write the header again.


hzeller:
This is my localized approach to solve the problem I encountered (`sf_write...()` failed after having manually called `sf_command(snd, SFC_UPDATE_HEADER_NOW,...)`. There might be better/more centralized ways to fix this, so I am open to implement it differently.

hzeller:
Alright, also added test that fails without the patch in the first submit and passes with the patch (and in a subsequent submit fixed that the test should only be run when FLAC is compiled in).

If you merge this via github, then the 'squash and merge' option is probably the best to only end up with a single commit.

erikd:
Merged manually. Thanks!

PR #308, opened 2017-08-08
Fix x86/x86_64 CPU detection regression

erikd:
Regression introduced in commit 8321896.
Reported by Anonymous Maarten.

Closes: https://github.com/erikd/libsndfile/pull/306


madebr:
Question related to this:
This Intel 64-bit cpu detection is only used in `src/sfendian.h` when using GCC for optimized byte 64-wapping.
Is the path ([source](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html)) that more optimized relative to __builtin_swapX or the byteswap.h header?

erikd:
If the `<byteswap.h>` header is present, then the built-ins are supposed to be used.

erikd:
@madebr I think that commit makes it a little more obvious that `<byteswap.h>` is preferred if available.

PR #314, opened 2017-09-04
Custom scale

eblanca:
During import, when converting from float to integer values, the caller can now specify a custom floating point value.
Old behavior is kept, and existing application won't break.

I figured out how to use that `SFE_BAD_COMMAND_PARAM` in case of wrong parameter, maybe that needs fixing...

### DIFF
--- a/src/sndfile.c
+++ b/src/sndfile.c
@@ -22,6 +22,7 @@
 #include	<string.h>
 #include	<ctype.h>
 #include	<assert.h>
+#include	<math.h>
 
 #include	"sndfile.h"
 #include	"sfendian.h"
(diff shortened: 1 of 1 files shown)


erikd:
This seems sane, but the formatting doesn't match the rest of the file and the documentation (in `doc/command.html`) needs updating.

eblanca:
Oh right, I will fix both.

BTW, in statement
`if (psf->float_int_mult && psf->float_max < 0.0)`
what's the purpose of that "psf->float_max < 0.0" test ?
I don't get it, maybe it checks for the very first setting of float_max (I see you init it with -1.0) ?

PR #315, opened 2017-09-04
Clipping of negative values

eblanca:
This PR fixes a couple of misplaced INT_MAX in floating point to integer conversion.
Or, at least, they seem so.

### DIFF
--- a/src/double64.c
+++ b/src/double64.c
@@ -519,7 +519,7 @@ d2i_clip_array (const double *src, int count, int *dest, double scale)
 
 		if (CPU_CLIPS_POSITIVE == 0 && tmp > (1.0 * INT_MAX))
 			dest [count] = INT_MAX ;
-		else if (CPU_CLIPS_NEGATIVE == 0 && tmp < (-1.0 * INT_MAX))
+		else if (CPU_CLIPS_NEGATIVE == 0 && tmp < (1.0 * INT_MIN))
 			dest [count] = INT_MIN ;
 		else
 			dest [count] = lrint (tmp) ;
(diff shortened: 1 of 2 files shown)


erikd:
I'm a bit concerned that there was no test that goes from failing to passing with this test. If you could add one, that would be awesome. If not I'll have to find some time to do it.

eblanca:
Any hint about how to build a meaningful new test?
I see there are lots of different files (of different types too) into tests/ , how do they work?

src/double64.c
```diff
@@ -519,7 +519,7 @@ d2i_clip_array (const double *src, int count, int *dest, double scale)
 
 		if (CPU_CLIPS_POSITIVE == 0 && tmp > (1.0 * INT_MAX))
 			dest [count] = INT_MAX ;
-		else if (CPU_CLIPS_NEGATIVE == 0 && tmp < (-1.0 * INT_MAX))
+		else if (CPU_CLIPS_NEGATIVE == 0 && tmp < (1.0 * INT_MIN))
```
erikd: The existing code was as expected. The reason is that the existing code was applying clipping symmetrically (same clip value for positive and negative values).

eblanca: If this was your purpose, then I think the statement
`dest [count] = INT_MIN ;`
should became
`dest [count] = -INT_MAX ;`
, should it ?

erikd: That may well be the case. I added an issue for it https://github.com/erikd/libsndfile/issues/324

erikd:
The code in this PR is actually not correct. There is an explicit reason for chosing `-1.0 * INIT_MAX` rather than `1.0 * INT_MAX` .

PR #316, opened 2017-09-08
fix clipping of negative values in flac format

### LINKED ISSUE #309: Clamping sample values for integer formats
I ran into an [issue](https://github.com/swh/timemachine/pull/4) when writing audio signals that contain floating point values greater 1.0 to integer formats with libsndfile. This leads to integer wrap-around in WAV and corrupted encoding in FLAC files. The values should be clamped to the maximum range of the target sample format. libsndfile is probably the right place to do this, not the app generating or recording the audio. What do you think?

jkroll20:
This fixes #309.


erikd:
Merged manually. Thanks.

Issue #320, opened 2017-09-19
psf_fseek behaves inconsistently when psf->fileoffset != 0


foolswood:
Ran into this when trying to get mp3 support going as it gets set by id3_skip when those tags are present (as they are in my test file).

I get behaviour (pseudocode for brevity) like:
```
seek(offset=0, whence=SEEK_CUR) -> 10
seek(offset=0, whence=SEEK_END) -> 8604437
seek(offset=0, whence=SEEK_CUR) -> 8438911
seek(offset=0, whence=SEEK_SET) -> 0
seek(offset=0, whence=SEEK_CUR) -> -165526
```

From code inspection I can see there's a bit of confusion about when we're talking about system file position and offset file position.

I'll make a branch and have a stab at tests and fixes, just thought I should document this somewhere.

foolswood:
Branch is: https://github.com/concert/libsndfile/tree/seek_fixes
Will put up a PR once I have fixes to go with the reproducing test.

erikd:
Closing this.

PR #321, opened 2017-09-19
Seeking with fileoffset fixes

foolswood:
I believe this tests and fixes the strange behaviour seen in #320.

Whether the fileoffset had been accounted for was unclear in the seek function and didn't seem to be fully correct. Hopefully this makes it clearer and avoids some of the edge cases.

Not sure whether I have handled pipeoffset correctly although I believe I have made explicit the existing behaviour.


foolswood:
I think this meets the style now, let me know if there are any other changes required.

erikd:
Merged manully. Thanks!

Issue #258, opened 2017-05-06
Add MP3 support


erikd:
The MP3 patents have now fully expired so libsndfile can now add support. 

The best bet would be https://www.mpg123.de which is LGPL and has updates as recently as this year.

I (@erikd) don't have sufficient time and motivation to do this, but would be happy to guide someone else.

evpobr:
Milestone?

erikd:
Until some actually steps forward to offer to do this and actually has some code that is worth reviewing, it doesn't make sense to have a milestone.

dabfalcon:
I hope someone steps forward and this feature would be great.

erikd:
It also seems that https://www.mpg123.de is a decoder only. All the other supported formats have both encode and decode so that would have to be the same for MP3 especially fore the CSound use case.

liblame may be the best bet.

evpobr:
Agreed, `mpg123` and `LAME` are the best.

Bad thing i will be buzy all the summer.

dabfalcon:
Bad thing I'm too n00b. How hard is it?
I'm good with PHP / web / did a course on C++.

erikd:
This is definitely more difficult than using libsndfile as a library, mainly because the libsndfile API was specifically designed to be easy to use. I would also rank this as more difficult than adding Opus support, because it requires you to hook up two libraries, one for decoding and one for encoding.

faroit:
Also interesting, from [this recent statement](http://www.audioblog.iis.fraunhofer.com/mp3-software-patents-licenses/):

> However, the end of the mp3 licensing program does not automatically mean that all mp3 technology is available license-free now. Apart from the core mp3 patents included in the licensing program, there might still be some implementation–specific patents (or patents for other functional enhancements) that have not expired. Thus, manufacturers will have to check the situation regarding their intended products first before including mp3.

jpffitch:
In csound at present we use a Russian library libmpadec for both read ad write.  Seems OK.Is it worth investigating?  I should add that it  has not been changed by the author for years but i have tweaked it, and have developer status for it.  Code in csound tree and on sourceforge

jpffitch:
Is there a template for adding a new format or does one just edit, say, ogg.c to new code?  Some time since I last tried (10 years) and things have changed.

erikd:
Closest thing to a template is `src/new.c`.

eblanca:
[...]

PR #322, opened 2017-09-20
[WIP] MP3 support

foolswood:
Putting this up early (in a distinctly unpolished state) to get some feedback about the overall shape and approach.

Status summary:
- [x] Read support
- [ ] Write support
- [ ] Make optional in build
- [x] dlopen lame


erikd:
> however in order to test it an example file to load is required. Is this acceptable, and if so where should this be in the source tree?

libsndfile has been around for nearly 20 years and has *never* had a binary file committed to the repository. i would like to keep it that way :). Hence, basics of read and write should be added in one go.

foolswood:
Fair enough, I'll have a stab at that. Did you have any comments on what's there so far?

src/mp3.c
```diff
@@ -227,16 +241,72 @@ mp3_read_2d	(SF_PRIVATE * psf, double * ptr, sf_count_t len)
...
+	char * err = NULL ;
+	while (lame_handle.dlhandle == NULL)
+	{	lame_handle.dlhandle = dlopen ("liblame.so", RTLD_LAZY) ;
+		if (lame_handle.dlhandle == NULL)
+			return dlerror () ;
+		lame_handle.init = dlsym (lame_handle.dlhandle, "lame_init") ;
+		if (lame_handle.init == NULL)
+		{	failed = 1 ;
+			break ;
+		}
```
erikd: Those braces don't match the coding style of the rest of the libsndfile code base,

The `Scripts/cstyle.py` script catches some, but not all issue.

src/mp3.c
```diff
@@ -0,0 +1,354 @@
...
+mp3_open (SF_PRIVATE * psf)
+{	psf->codec_data = NULL ;
+	switch (psf->file.mode) {
+	case (SFM_READ) :
+		return mp3_read_open (psf) ;
+	case (SFM_WRITE) :
+		return mp3_write_open (psf) ;
+	default:
+		return SFE_UNIMPLEMENTED ;
+	}
```
erikd: The libsndfile codebase has an coding style which it adheres to with a reasonably high level of consistency.

The is an a Python script `Scripts/cstyle.py` which catchs a large number of issues. You should still have a look at the other code and match the style.

src/mp3.c
```diff
@@ -0,0 +1,354 @@
...
+*/
+
+// FIXME: use mpg123 error string reporting
+static int
+mp3_read_open (SF_PRIVATE * psf)
+{
+	int decoder_err = MPG123_OK ;
+	mpg123_handle * decoder ;
+	if (mpg123_initialised == 0)
+	{	int decoder_init_err = mpg123_init () ;
```
erikd: Can `mpg123_init()` not be called multiple times without blowing up?

src/mp3.c
```diff
@@ -0,0 +1,354 @@
...
+	psf->read_double = mp3_read_2d ;
+
+	return 0 ;
+}
+
+static int
+mp3_read_close (SF_PRIVATE * psf)
+{	mpg123_handle * decoder = psf->codec_data ;
+	if (decoder != NULL)
+	{	// mpg123_close(decoder); <- Not sure if we need this?
```
erikd: If it doesn't blow up, call it anyway.

src/mp3.c
```diff
@@ -0,0 +1,354 @@
...
+		mpg123_delete (decoder) ;
+		// The psf sometimes gets reused:
+		psf->codec_data = NULL ;
+	}
+	if (!--mpg123_initialised)
+		mpg123_exit () ;
```
erikd: This is kinda weird! What happens without that `mpg123_exit()` call? Does it leak memory?

src/mp3.c
foolswood: IIRC it did, which is why it was commented out, but I'll check.

src/mp3.c
foolswood: I thought I had got it all past cstyle.py (which I believe is triggered in the pre-commit hook), but I'll check elsewhere for the preferred bracket style.

src/mp3.c
foolswood: The `mpg123_init()` and `mpg123_exit()` calls are following the canonical documented use of the library. I've looked through the source, which suggests there is some cheating to be had and contacted the mpg123 developers to see what their future plans for the methods are.

src/mp3.c
erikd: I've just added some new rules to the `Script/cstyle.py` script that should help with some more coding style issues. The one that's difficult to add to that script is the closing curly indentation which the current source code uses. It should be like:
```
    if (whatever)
    {   do_soometing ;
        } ;
```
Notice how the closing curly is indented one more level than the open curly brace.

foolswood:
@diemoschwarz I had dispaired of this somewhat (see the discussion here and on #258 for why). IIRC (which I might not because this was 18 months ago) I did have playback working at one point (although I can't remember under what conditions).

evpobr:
#499 is almost ready, closing this one.

PR #325, opened 2017-09-27
sfe_copy_data_fp: check value of "max" variable for being normal

### LINKED ISSUE #317: type confusion(double's NAN & INF) vuls was found in function sfe_copy_data_fp()
Some type confusion vuls was found in function sfe_copy_data_fp() in programs/common.c of libsndfile-1.0.28:
sfe_copy_data_fp (SNDFILE *outfile, SNDFILE *infile, int channels, int normalize)
{	static double	data [BUFFER_LEN], max ;
	int		frames, readcount, k ;

	frames = BUFFER_LEN / channels ;
	readcount = frames ;

	sf_command (infile, SFC_CALC_SIGNAL_MAX, &max, sizeof (max)) ;

	if (!normalize && max < 1.0)
	{	while (readcount > 0)
		{	readcount = sf_readf_double (infile, data, frames) ;
			sf_writef_double (outfile, data, readcount) ;
			} ;
		}
	else
	{	sf_command (infile, SFC_SET_NORM_DOUBLE, NULL, SF_FALSE) ;

		while (readcount > 0)
		{	readcount = sf_readf_double (infile, data, frames) ;//can read from host_read_d(), so data can be controled by user.
			for (k = 0 ; k < readcount * channels ; k++)
				data [k] /= max ;
			sf_writef_double (outfile, data, readcount) ;
			} ;
	} ;

	return ;
} /* sfe_copy_data_fp */

 There are two ways to generate NAN or INF double type by a craft audio file. first, the value of max may be 0 or inf, so it will  subsequently get a NAN while "data[k]/=max"; second , the data[] may came from audio file directly, so data[k] may be NAN or INF and so does "data [k] /= max".
Then, while d2ulaw_array()/f2ulaw_array()/d2alaw_array()/f2alaw_array() in ulaw.c and alaw.c, a SEGV will appear:

$ gdb programs/sndfile-convert
GNU gdb (GDB) 7.9
Copyright (C) 2015 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-unknown-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from programs/sndfile-convert...done.
(gdb) run -ulaw ../../libsndfile-1.0.28/fuzz/crash1-get-nan-from-host xxx.vox
Starting program: /home/jiangxin/experiment/fuzz/AFL/target/debug/libsndfile-1.0.28/programs/sndfile-convert -ulaw ../../libsndfile-1.0.28/fuzz/crash1-get-nan-from-host xxx.vox

Program received signal SIGSEGV, Segmentation fault.
0x0000000000453781 in d2ulaw_array (ptr=0x68ba20 <data>, count=4, buffer=0x7fffffffc010 "1\005\352\177J\377\200\200\200\200\377\377", 'J' <repeats 12 times>, "1\261\061\061\061\061\061\061", normfact=8191.75) at ulaw.c:853
853				buffer [count] = 0x7F & ulaw_encode [- lrint (normfact * ptr [count])] ;
(gdb) bt
#0  0x0000000000453781 in d2ulaw_array (ptr=0x68ba20 <data>, count=4, buffer=0x7fffffffc010 "1\005\352\177J\377\200\200\200\200\377\377", 'J' <repeats 12 times>, "1\261\061\061\061\061\061\061", normfact=8191.75) at ulaw.c:853
#1  0x0000000000453e89 in ulaw_write_d2ulaw (psf=0x69d2f0, ptr=0x68ba20 <data>, len=12) at ulaw.c:1041
#2  0x0000000000409a7c in sf_writef_double (sndfile=0x69d2f0, ptr=0x68ba20 <data>, frames=12) at sndfile.c:2570
#3  0x00000000004024a1 in sfe_copy_data_fp (outfile=0x69d2f0, infile=0x69a010, channels=1, normalize=0) at common.c:79
#4  0x0000000000402122 in main (argc=4, argv=0x7fffffffe228) at sndfile-convert.c:338
(gdb) p ptr[count]
$1 = nan(0xd40a620000000)
(gdb) i r
rax            0x8000000000000000	-9223372036854775808
rbx            0x7fffffffc014	140737488338964
rcx            0xc	12
rdx            0x20	32
rsi            0xc	12
rdi            0x68ba20	6863392
rbp            0x7fffffffbfe0	0x7fffffffbfe0
rsp            0x7fffffffbfb0	0x7fffffffbfb0
r8             0x0	0
r9             0x100000	1048576
r10            0x0	0
r11            0x246	582
r12            0x4015a0	4199840
r13            0x7fffffffe220	140737488347680
r14            0x0	0
r15            0x0	0
rip            0x453781	0x453781 <d2ulaw_array+174>
eflags         0x10a87	[ CF PF SF IF OF RF ]
cs             0x33	51
ss             0x2b	43
ds             0x0	0
es             0x0	0
fs             0x0	0
gs             0x0	0
(gdb) x/i $pc
=> 0x453781 <d2ulaw_array+174>:	movzbl 0x689760(%rax),%eax
(gdb) detach
Detaching from program: /home/jiangxin/experiment/fuzz/AFL/target/debug/libsndfile-1.0.28/programs/sndfile-convert, process 14340
(gdb) q

fabiangreffrath:
and check elements of the data[] array for being finite.

Both checks use functions provided by the <math.h> header as declared
by the C99 standard.

Fixes #317
CVE-2017-14245
CVE-2017-14246

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -496,6 +496,9 @@ if (BUILD_PROGRAMS)
 		programs/common.h)
 	add_executable (sndfile-convert ${sndfile_convert_SOURCES})
 	target_link_libraries(sndfile-convert PUBLIC ${SNDFILE_TARGET})
+	if (LIBM_REQUIRED)
+		target_link_libraries(sndfile-convert PRIVATE ${M_LIBRARY})
+	endif ()
 
 # sndfile-cmp
 
@@ -514,6 +517,9 @@ if (BUILD_PROGRAMS)
 		programs/common.h)
 	add_executable (sndfile-metadata-set ${sndfile_metadata_set_SOURCES})
 	target_link_libraries(sndfile-metadata-set PUBLIC ${SNDFILE_TARGET})
+	if (LIBM_REQUIRED)
+		target_link_libraries(sndfile-metadata-set PRIVATE ${M_LIBRARY})
+	endif ()
 
 # sndfile-metadata-get
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -477,7 +477,7 @@ if (BUILD_PROGRAMS)
 	add_executable (sndfile-play ${sndfile_play_SOURCES})
 	target_link_libraries(sndfile-play PUBLIC ${SNDFILE_TARGET})
 	if (LIBM_REQUIRED)
-		target_link_libraries(sndfile-convert PRIVATE ${M_LIBRARY})
+		target_link_libraries(sndfile-play PRIVATE ${M_LIBRARY})
 	endif ()
 	if (WIN32)
 		target_link_libraries(sndfile-play PRIVATE Winmm.lib)
@@ -512,7 +512,7 @@ if (BUILD_PROGRAMS)
 	add_executable (sndfile-cmp ${sndfile_cmp_SOURCES})
 	target_link_libraries(sndfile-cmp PUBLIC ${SNDFILE_TARGET})
 	if (LIBM_REQUIRED)
-		target_link_libraries(sndfile-convert PRIVATE ${M_LIBRARY})
+		target_link_libraries(sndfile-cmp PRIVATE ${M_LIBRARY})
 	endif ()
 
 # sndfile-metadata-set
@@ -536,7 +536,7 @@ if (BUILD_PROGRAMS)
 	add_executable (sndfile-metadata-get ${sndfile_metadata_get_SOURCES})
 	target_link_libraries(sndfile-metadata-get PUBLIC ${SNDFILE_TARGET})
 	if (LIBM_REQUIRED)
-		target_link_libraries(sndfile-convert PRIVATE ${M_LIBRARY})
+		target_link_libraries(sndfile-metadata-get PRIVATE ${M_LIBRARY})
 	endif ()
 
 # sndfile-interleave
@@ -548,7 +548,7 @@ if (BUILD_PROGRAMS)
 	add_executable (sndfile-interleave ${sndfile_interleave_SOURCES})
 	target_link_libraries(sndfile-interleave PUBLIC ${SNDFILE_TARGET})
 	if (LIBM_REQUIRED)
-		target_link_libraries(sndfile-convert PRIVATE ${M_LIBRARY})
+		target_link_libraries(sndfile-interleave PRIVATE ${M_LIBRARY})
 	endif ()
 
 # sndfile-deinterleave
@@ -560,7 +560,7 @@ if (BUILD_PROGRAMS)
 	add_executable (sndfile-deinterleave ${sndfile_deinterleave_SOURCES})
 	target_link_libraries(sndfile-deinterleave PUBLIC ${SNDFILE_TARGET})
 	if (LIBM_REQUIRED)
-		target_link_libraries(sndfile-convert PRIVATE ${M_LIBRARY})
+		target_link_libraries(sndfile-deinterleave PRIVATE ${M_LIBRARY})
 	endif ()
 
 # sndfile-concat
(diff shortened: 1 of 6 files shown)


fabiangreffrath:
Apparently we need to add `-lm` to the linker flags for this to work out as expected.

fabiangreffrath:
> Apparently we need to add -lm to the linker flags for this to work out as expected.

That being said, my automake foo is bad and my cmake foo is even worse, so I'd appreciate if someone else could take care of that and add a check to add `-lm` to the linker flags, thanks!

evpobr:
CMake doesn't work? It performs `m` library check:

```cmake
find_library (M_LIBRARY m)
if (M_LIBRARY)
	# Check if he need to link 'm' for math functions
	check_library_exists (m floor "" LIBM_REQUIRED)
	if (LIBM_REQUIRED)
		list (APPEND CMAKE_REQUIRED_LIBRARIES ${M_LIBRARY})
	else ()
		unset (M_LIBRARY)
	endif ()
endif ()
mark_as_advanced (M_LIBRARY)
```

fabiangreffrath:
Hm, I don't understand why the travis check fails, then? Or, why it doesn't fail earlier, when other objects using math library functions are linked?

fabiangreffrath:
I see, only select programs (i.e. sndfile-info) are linked against `M_LIBRARY` as needed. But since I added new math functions to `common.c` which is linked by all of the programs, I will have to add
```
	if (BUILD_SHARED_LIBS AND LIBM_REQUIRED)
		target_link_libraries(program PRIVATE ${M_LIBRARY})
	endif ()
```
for each program? Or is there a way in cmake to tell "link any object against `-lm` that also links against `common.c`"?

evpobr:
Yes atfer your changes we need to add:

```cmake
if (LIBM_REQUIRED)
        target_link_libraries(<program-name> PRIVATE ${M_LIBRARY})
endif ()
```

to every program that calls `sfe_copy_data_fp`.

`BUILD_SHARED_LIBS` must be removed i guess.

evpobr:
For Autotools look at this:

```c
src_libsndfile_la_LIBADD = src/GSM610/libgsm.la src/G72x/libg72x.la src/ALAC/libalac.la \
	src/libcommon.la $(EXTERNAL_XIPH_LIBS) -lm
```

Try to add `-lm` to programs `LDADD` variable:

```c
programs_sndfile_info_LDADD = src/libsndfile.la -lm
```

To make it better: i think sometimes `m` is not required and not present, so adding `-lm` will fail. I found macro to check it. Add this to `configure.ac`:

```m4
AC_CHECK_LIBM
AC_SUBST(LIBM)
```
This macro will check if `m` is required and set `LIBM` variable to the name of `m` library. Then in `Makefile.am` use `$(LIBM)` variable instead of `-lm`.

```
src_libsndfile_la_LIBADD = src/GSM610/libgsm.la src/G72x/libg72x.la src/ALAC/libalac.la \
	src/libcommon.la $(EXTERNAL_XIPH_LIBS) $(LIBM)
```

Now if no `m` found, $(LIBM) will be empty and will not harm.

fabiangreffrath:
Hm, no, apparently any program linking in `common.o` also needs to link against `-lm`, regardless of calling `sfe_copy_data_fp()` or not.

fabiangreffrath:
Now the checks pass! \o/

erikd:
I thought I commented on this but either I screwed up or Github lost my comment.

The lost comment was saying that I don't think this is the right solution. The problem should be fixed in the library itself, rather than in (I suspect) a  utility program that often isn't even installed. If its difxed in the library then all programs that link against libsndfile benefit, not just this one utility program.

evpobr:
I'll check it.

PR #328, opened 2017-10-01
Remove unused internal header includes from examples.

dbgasaway:
These headers aren't used by the examples (as far as I could tell). Removing these includes would help make them look more like real user programs that may not have access to these files and make it faster and require less extra files to compile after copy-pasting somewhere else.


erikd:
Thanks,

PR #329, opened 2017-10-08
Make git checkout buildable out-of-source

SoapGentoo:
* Currently, libsndfile requires in-tree bootstrapping
  and creation of the tarball. For Gentoo, some users
  might want to build the latest git checkout. This
  commit makes out-of-source building of all the
  necessary dependencies possible.


erikd:
Tweaked the comment message and merged,

Thanks.

PR #334, opened 2017-10-21
src/sfconfig.h: Fix CPU architecture detection under Windows

evpobr:
`__x86_64__`, `__i486__`, `__i586__`, `__i686__` are not defined under MSVC, so `CPU_IS_X86` and `CPU_IS_X86_64` are always `0`:

default

[MSDN: Predefined Macros](https://msdn.microsoft.com/en-US/library/b0084kay(v=vs.140).aspx)

### DIFF
--- a/src/sfconfig.h
+++ b/src/sfconfig.h
@@ -113,10 +113,10 @@
 #define HAVE_X86INTRIN_H 0
 #endif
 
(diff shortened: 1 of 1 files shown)


erikd:
Thanks!

PR #343, opened 2017-12-04
[WIP] Documentation updates

evpobr:


evpobr:
From here are missing API functions and command are documented.

evpobr:
Finished this part.

@erikd, i'm thinking about splitting whole API reference to small chunks, one file per function, struct etc. Also i want add sections to topics. Something like this:

# sf_open

Open sound file from path.

## Synopsis

```c
#include <sndfile.h>

SNDFILE*  sf_open    (const char *path, int mode, SF_INFO *sfinfo) ;
```
## Description

The `sf_open()` function opens the sound file at the specified path. The filename is byte encoded, but may be utf-8 on Linux, while on Mac OS X it will use the filesystem character set. On Windows, there is also a Windows specific sf_wchar_open() that takes a UTF16_BE encoded filename.

...

## Return value

On success, the sf_open function returns a non-NULL pointer which should be passed as the first parameter to all subsequent libsndfile calls dealing with that audio file. On fail, the sf_open function returns a NULL pointer. An explanation of the error can obtained by passing NULL to sf_strerror.

## See also

* SF_INFO
* sf_open_fd, sf_open_virtual
* sf_close

evpobr:
Also i think it is time to switch from HTML as source to another format. As output format HTML is fine, but it is not flexible enough.

I prefer some XML-based format, DocBook or DITA.

Something like MarkDown is also OK, much easier to write. Maybe AsciiDoc or RestructuredText. MarkDown is too simple, i tried it.

Doxygen... Nice to use, header is autodocumented. But output HTML is completely crappy.

janstary:
@evpobr I rewrote the manages of the sndfile-*(1) utils into mdoc(7) some time ago,
and it has been on m TODO list since then to also rewrite the API doc.

I do believe mdoc(7) is the standard, right way to do it; it has been for decades.
I agree HTML is a poor choice; please don't even mention XML or Markdown.

evpobr:
@janstary `mdoc` is so popular standard i cannot even find an article on Wikipedia. You talk about that FreeBSD utility? What about WIndows?

Also make it clear for me - why not ReStructuredText, AsciiDoc or DocBook? We both know Markdown is not right choce for documentation.

Anyway, you are free to try, let's look at this.

janstary:
[mdoc(7)](http://man.openbsd.org/mdoc#HISTORY) is a language as old as 4.4BSD.
Look at the manpage of, say, ssh(1). That's the language it's written in.
See the excelent [History of UNIX manopges](http://manpages.bsd.lv/history.html)

janstary:
Tools to process this language have been around for decades: mostly `groff` (on most Linuxes),
`mandoc` relatively recently (but still it's years now) on the BSDs, Heirloom on more exotic systems.
There is nothing to install, tha language is ubiquituous. Increasingly, even MacOSX converges to it,
taking parts from FreeBSD.

The main advantages of this _language_ over the other possibilities are such
that I won't even discuss them in the Markdown text window :-)
I will cook up an example and post it to the mailing list.

Also, this discussion lagely happend already when Erik accepted the mdoc(7) rewrite
of the `*.1` manpages in https://github.com/erikd/libsndfile/commit/d4e014e82d92f41c91a53df33be4d8914047b1b8

janstary:
See also the general comments on documentation languages and tool in https://undeadly.org/cgi?action=article&sid=20170304230520

janstary:
https://github.com/janstary/libsndfile/tree/apidoc/doc
This is the branch where I will push my idea of the api doc.
So far, there is `sf_open(3)` as an example.

erikd:
@janstary Are there tools for converting `mdoc` to HTML?

janstary:
Looking at a few html manpages produced by groff and mandoc,
I think the mandoc -Thtml output is considerably better.
On the other hand, groff is ubiquituous.

erikd:
@evpobr Are `mdoc` and `mandoc` available on Windows?

janstary:
`mdoc` is a language. Is `html` available on Windows?

`mandoc` is available anywhere I have tried, but I don't do Windows, sorry.

Issue #345, opened 2017-12-11
Is SFC_GET_LOG_INFO documentation wrong ?


aybe:
When I call that command with a correctly opened stream, I get an empty string which obviously makes sense to me.

It's only when I open a non-existent path, so the call fails, and then pass a null SNDFILE pointer to the command that I get a log string.

Did I miss something or documentation is wrong ?

erikd:
Passing a NULL pointer as the SNDFILE only works the `sf_open` fails.

If `sf_open` returns non-NULL, you should use that pointer in the call to `sf_command`.

aybe:
Yes, I understand that, what I meant was:

When there is a failure I know about the cause thanks to the log.

But when the file opens, the log is empty. Or, said differently : I expected something like 'file opened successfully, format is ... etc'.

Hope that makes sense to you :)

erikd:
Your code should look something like:
```
  	sndfile = sf_open (pathname, SFM_READ, &info) ;
	sf_command (sndfile, SFC_GET_LOG_INFO, logbuf, sizeof (logbuf)) ;
        puts (logbuf) ;
        sf_close (sndfile) ;
```

evpobr:
> This log buffer can often contain a good reason for why libsndfile failed to open a particular file.

Documentation is correct, log is not always contains information.

> But when the file opens, the log is empty. Or, said differently : I expected something like 'file opened successfully, format is ... etc'.

Can you make pull request? We really need help.

erikd:
We only need a PR if there is actually something wrong with the documentation. That has not yet been confirmed.

aybe:
I've found the error !

In http://www.mega-nerd.com/libsndfile/command.html#SFC_GET_LOG_INFO :

```
Return value:
This call will return the length of the retrieved version string.
```
This is wrong, by debugging https://github.com/erikd/libsndfile/blob/master/src/sndfile.c#L1033 as I understand C by default would return last value in case of no `return`, but in fact there is https://github.com/erikd/libsndfile/blob/master/src/sndfile.c#L1373 and effectively we get `0`.

And as I used to marshal the string from C to C# it was always empty, now I use a method similar to yours with `puts` and effectively things work as expected:

For a file that exists:

```
File : 'C:\Users\Aybe\Music\Forme - SÃ¶ndag.wav' (utf-8 converted from ucs-2)
Length : 53048084
RIFF : 53048076
WAVE
fmt  : 16
[...]

PR #349, opened 2017-12-16
src/sndfile.c: Set SFC_GET_LIB_VERSION return result to 0 on error

evpobr:
If `data == NULL` string is not copied but command returns `SFE_BAD_COMMAND_PARAM (31)`. It can be misunsderstood as string length.

Related to issue #345.

### DIFF
--- a/src/sndfile.c
+++ b/src/sndfile.c
@@ -888,7 +888,7 @@ sf_command	(SNDFILE *sndfile, int command, void *data, int datasize)
 			if (data == NULL)
 			{	if (psf)
 					psf->error = SFE_BAD_COMMAND_PARAM ;
-				return SFE_BAD_COMMAND_PARAM ;
+				return 0 ;
 				} ;
 			snprintf (data, datasize, "%s", sf_version_string ()) ;
 			return strlen (data) ;


src/sndfile.c
```diff
@@ -888,7 +888,7 @@ sf_command	(SNDFILE *sndfile, int command, void *data, int datasize)
 			if (data == NULL)
 			{	if (psf)
 					psf->error = SFE_BAD_COMMAND_PARAM ;
-				return SFE_BAD_COMMAND_PARAM ;
+				return 0 ;
```
erikd: Sorry, that's not correct. Calling code should check the return of `sf_command` and only if it is non-zero should the data be trusted.

evpobr: Documentation says that `SFC_GET_LIB_VERSION returns string length.`But it is not.

The problem is mixing string length and error code at return.

Error return result was 31 (SFE_BAD_COMMAND_PARAM) before, that can be interpreted as success with string of 31 chars returned.

Now it is zero and clearly indicates that nothing was copied.

erikd: > Documentation says that SFC_GET_LIB_VERSION returns string length.`But it is not.

The code is correct, the documentation needs to be fixed.

erikd: On second thought, this probably is a better solution.

PR #353, opened 2018-01-05
unbreak sed(1) in pre-commit

janstary:
This is MacOS 10.13.2

Running `git commit -a` currently fails in pre-commit with

```
sed: illegal option -- -
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

The sed dance results in `sed -r` wrongly,
not to mention that `sed --version` already assumes it _is_ GNU.

In fact, GNU, BSD and MacOS all have `-E`, altough GNU does not document it.
Feel free to remove the comments. Move it to where it's needed, while here.

This PR fixes the immediate fail for me and works on MacOS, OpenBSD and Linux.
I would like to get rid of the sed-and-head-and-grepawkery altogether,
but I am not sure what exactly it's inteded to do.

### DIFF
--- a/Scripts/git-pre-commit-hook
+++ b/Scripts/git-pre-commit-hook
@@ -1,5 +1,4 @@
 #!/bin/sh
-#
 
 if git rev-parse --verify HEAD >/dev/null 2>&1 ; then
 	against=HEAD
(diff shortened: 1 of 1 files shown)


erikd:
Applied. Thanks!

PR #354, opened 2018-01-28
remove BEOS support

janstary:
Does anyone use sndfile on BEOS?
Does anyone use BEOS?
Does BEOS exist?


erikd:
Merged manually. Thanks!

PR #355, opened 2018-01-28
drop MacOSX from sndfile-play

janstary:
We don't play on MacOSX >= 10.8 (the code for the new API is not written),
and support for MacOSX older than that was removed some time ago.
So drop the last remnants; we just don't play on MacOSX.


erikd:
Thanks!

Issue #43, opened 2013-06-15
Improve error reporting of sndfile-convert


erikd:
The command:

```
sndfile-convert /tmp/a.ogg /tmp/a.wav
Error : output file format is invalid (0x00010060).
```

should provide a better error messages.

The problem is that the above command is trying to convert a file containing Vorbis data in an Ogg container into a file containing Vorbis data in a WAV container, but the WAV container doesn't support Vorbis data.

Typically what is needed is something like:

```
sndfile-convert -pcm16 /tmp/a.ogg /tmp/a.wav
```

to convert the audio data into a format that is accepted by the the WAV container.

PR #357, opened 2018-01-31
Check if conversion failed due to no. of channels

jwakely:
This fixes the problem noted in #43 where attempting to convert a stereo WAV to SDS gives a misleading error about the codec, when the real problem is that the output format only supports mono not stereo.

### DIFF
--- a/programs/sndfile-convert.c
+++ b/programs/sndfile-convert.c
@@ -114,6 +114,13 @@ report_format_error_exit (const char * argv0, SF_INFO * sfinfo)
 		exit (1) ;
 		} ;
 
+	int channels = sfinfo->channels;
+	sfinfo->channels = 1;
+	if (sf_format_check (sfinfo))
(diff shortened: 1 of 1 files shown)


jwakely:
This passes all exising tests. I didn't add a new test or modify any to check for the new error message - should I have done?

programs/sndfile-convert.c
```diff
@@ -114,6 +114,13 @@ report_format_error_exit (const char * argv0, SF_INFO * sfinfo)
 		exit (1) ;
 		} ;
 
+	int channels = sfinfo->channels;
```
erikd: This fails in travis. See https://travis-ci.org/erikd/libsndfile/jobs/335882223

That variable declaration will need to go at the top of the block.

erikd:
Nevermind. Fixed in new PR #359 .

erikd:
Officially libsndfile is C99, but I don't like to unnecessarily break non-C99 compilers.

PR #361, opened 2018-02-07
Support for NMS ADPCM in WAV files

arthurt:
This pull request add support for NMS ADPCM audio inside WAV files.

If files created from VoxStudio are to be believed, WAV format 0x0038 (NMS_VBXADPCM) does indicate a NMS ADPCM payload. A bitwidth field of 2, 3 or 4 indicates the 16kbps, 24kbps and 32kbps variants respectively. Other bitwidth values /might/ exist, including a varying bitwidth form, but these are not supported.

This pull request adds support for SF_FORMAT_WAV | SF_FORMAT_NMS_ADPCM_* as well as lossy_comp_test test for NMS ADPCM using the WAV container.


erikd:
Very cool! Probably won't get a chance to look at this properly until the weekend.

erikd:
Thanks for this and sorry for the delay.

PR #383, opened 2018-03-19
Portable suffix rules

SoapGentoo:
@janstary @erikd
This is supposed to make libsndfile work with BSD make. I have tried it using NetBSD's make in Gentoo, and after adding lots of awful crutches because `$<` is not portable on BSD make, I managed to get it working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,4 @@ Debug/
 Release/
 bin/
 obj/
+octave-workspace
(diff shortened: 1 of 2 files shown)


erikd:
I'm fine with this. Waiting for @janstary .

janstary:
Works for me

erikd:
Oh, I squashed that iand did PR #385 which I just merged.

Thanks anyway!

Issue #110, opened 2015-12-10
Read/write Ogg Vorbis to pipe?


keidax:
I'm having some trouble reading a Vorbis stream from a pipe. I get the error message `Supported file format but file is malformed.` when calling `sf_open_fd` on the pipe.

Output of  `sf_command(SFC_GET_LOG_INFO)`:
```
Length : unknown
Ogg stream data : Vorbis
Stream serialno : 0679846748
Vorbis library version : Xiph.Org libVorbis 1.3.5
Input does not appear to be an Ogg bitstream.
```

I can write the pipe contents to a regular file and read from that without any trouble, but I'd like to be able to do this in a continuous stream. Do libsndfile support reading or writing Vorbis to/from a non-seekable pipe, FIFO, or something similar? If not, a warning message or note in the doumentation would be nice.

erikd:
It doesn't currently. It should be possible to do it but its not high on my list of priorities.

rsalinas:
I am also interested in reading an OGG stream.  Apparently it tries to `fseek` and fails.  How far away is _libsndfile_ from doing that? I saw a few `seeks` in a quick look.  If you have any helping tips, I may try to fix the problem myself, thanks!

erikd:
Sorry, I don't have any tips to help with this issue. Not even 100% sure its possible,

arthurt:
Looked into this, and it is fixable. The reason Ogg encapsulated files currently require a seek is because the container code in src/ogg.c loads the first packet of the first stream of the first page of the file, in order that it might detect what payload is present. It then resets the state by clearing all the Ogg structures, and trying to seek to the beginning in order that the codec code (/src/ogg_vorbis.c, etc) can start the process anew.

I think this is from the time when the only supported Ogg format was Vorbis, as src/ogg.c would seem to be a version of src/ogg_vorbis.c repurposed to deciding what codec to open.

A fix could be, once a packet has been loaded and identified as containing a supported stream, leaving the Ogg data structures (sync, page, stream, packet) initialized and loaded, and modify the ogg_* codecs to expect or check for that upon opening.

erikd:
This should be fixed now. Can someone please confirm?

c776617:
I for one can confirm that reading an OGG Vorbis file using `sf_open_virtual` works now and did not in 1.0.28. (tested on Win64) Same for FLAC.

Here's the `streamsndfile.cpp` test code:

```
/*
** streamsndfile by c776617
[...]

PR #384, opened 2018-03-19
Avoid seeks in ogg_open (), Introduce ogg_read_first_page ()

arthurt:
This pull request merges duplicate code from from src/ogg.c and src/ogg_vorbis.c into a common function.. The code in question is for reading the first page of an Ogg bitstream. The benefits of such a merge are:

1) Each codec encapsulated by Ogg requires the functionality, so future codecs may use it without duplication as well.

2) Removal of the duplicate code from src/ogg_vorbis.c allows for a more structured way for calling ogg_read_first_page () only as needed, remove the need for re-reading the beginning of the file. In my tests this fixes issue #110 .

This pull request also changes the way ogg_vorbis_open () and ogg_open () interact. Previously ogg_vorbis_open () assumed nothing about the state of the container state, re-initializing all Ogg related structures itself and assuming that the read position was the start of the file. This required ogg_open () to seek to the beginning of the stream before calling ogg_vorbis_open (). After this change ogg_vorbis_open () will assumesthat the container state is already initialized, with the first page loaded.

On backwards seek, src/ogg_vorbis.c still has to re-read the first page, and it will call ogg_read_first_page appropriately.

FLAC in ogg behaviour is unaffected by this change. Speex in ogg (src/ogg_speex.c) has not been modified to take advantage of ogg_read_first_page (), and should be unaffected as well.

### DIFF
--- a/src/ogg.c
+++ b/src/ogg.c
@@ -46,6 +46,87 @@ static int	ogg_close (SF_PRIVATE *psf) ;
 static int	ogg_stream_classify (SF_PRIVATE *psf, OGG_PRIVATE * odata) ;
 static int	ogg_page_classify (SF_PRIVATE * psf, const ogg_page * og) ;
 
+int ogg_read_first_page (SF_PRIVATE *psf, OGG_PRIVATE *odata)
+{	char *buffer ;
+	int	bytes ;
+
+	/*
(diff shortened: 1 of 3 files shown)


erikd:
Thanks! Does that mean that #110 can be closed?

PR #394, opened 2018-05-13
Treat sf_open() path parameter as UTF-8 sequence

evpobr:


doc/api.html
```diff
@@ -129,6 +129,12 @@ <H2><B>File Open Function</B></H2>
 On Windows, there is also a Windows specific sf_wchar_open() that takes a
 UTF16_BE encoded filename.
 </P>
+<blockquote>
+<p><b>Note:</b></p>
+<p>Starting from 1.0.29 release, behavior of <code>sf_open()</code> on Windows
+ platform was changed, <code>path</code> parameter is treated as UTF-8 string.
+</p>
```
erikd: As a native English speaker, that test is just a tiny bit off. Can we change it to:
```
From the 1.0.29 release, the behavior of the <code>sf_open()</code>function on Windows
platform has changed, with the <code>path</code> parameter being treated as a UTF-8 string.
```

src/sndfile.c
```diff
@@ -328,6 +332,18 @@ sf_open	(const char *path, int mode, SF_INFO *sfinfo)
 	/* Ultimate sanity check. */
 	assert (sizeof (sf_count_t) == 8) ;
 
+#if OS_IS_WIN32
+
+	wchar_t wpath [SF_FILENAME_LEN] ;
+
+	int dwRet = MultiByteToWideChar (CP_UTF8, MB_ERR_INVALID_CHARS, path, -1,
```
erikd: I don't see a good reason to line wrap this.

src/sndfile.c
```diff
@@ -328,6 +332,18 @@ sf_open	(const char *path, int mode, SF_INFO *sfinfo)
...
+		wpath, SF_FILENAME_LEN );
```
erikd: and the `cstyle.py` script should probably have complained about that.

erikd:
Thanks @evpobr !

PR #400, opened 2018-08-01
Build with Selectable External Libraries, Rather Than All-Or-Nothing

arthurt:
Changes to both GNU autotools and CMake build systems to allow for individually selectable external libraries (not just Xiph.org ones too). Also changes config.h definitions, removing HAVE_EXTERNAL_XIPH_LIBS and introducing HAVE_OGG, HAVE_VORBIS, HAVE_FLAC and HAVE_SPEEX.

This allows for greater flexibility of which Xiph.org codecs are enabled for a build, and eases possible future work of introducing new external library dependencies.

The impetus for this change is for allowing easier Opus integration, and future MP3 integration.

Also removes linking to the (unused) libvorbisfile from the CMake build scripts.


evpobr:
@erikd, we can check with all external libraries only. Check individual configurations is just waste of time.

arthurt:
> How are we going to test all the combinations? Just with Ogg, Vorbis and FLAC there are 6 combinations, some of which are not valid, eg Vorbis without Ogg is not a valid configuration.

Both build scripts make Vorbis (and Speex) support dependent on having Ogg, so the invalid combinations are not possible. Further, in the autotools configure script if a user explicitly demands Vorbis (or Speex) when Ogg isn't available, a warning is printed. Likewise, for trying to explicitly enable Speex when experimental isn't enabled.

In the code I have used (HAVE_VORBIS && HAVE_OGG) to be explicit, but HAVE_VORBIS would be sufficient, as HAVE_VORBIS implies HAVE_OGG.

Echoing evpobr, is auto-testing all individual combinations worth it? I have of course tested manually when writing this.

>> The impetus for this change is for allowing easier Opus integration, and future MP3 integration.

> I don't see how this make it easier.

For Opus support, no, it's not more work to go with the status quo, although it does force users to have all of Voribs, FLAC, and Opus. For MP3 support I think it becomes unruly to expect all Xiph.org and a non-Xiph.org MP3 library as an all-or-nothing thing. Maybe a user has no MP3 library but wants Ogg/Vorbis support. As such it makes sense to me to split MP3 external dependencies from the Xiph.org ones, and then it isn't that much more work to split the Xiph.org ones up too.

I'm currently reworking my Opus tree for pulling. I thought that this change would be nice to have first. I've also started on MP3 support with liblame (decode too) for which it would be nice to have this change.

The Travis builds on Linux are currently failing on autotools. Would appear aclocal cannot find pkgconfig's pkg.m4, which makes no sense to me as the failing line is un-altered, and, of course, "it works for me." Travis' Apple build succeeds with autotools.

erikd:
So, the valid combinations (ignoring Speex because its not actually working yet) are:

*  No external libs
* FLAC only
* FLAC with Ogg
* Ogg/Vorbis only
* FLAC and Ogg/Vorbis

That's current 5 configurations, all of which requires separate testing. Each time we add a new codec, the number of combinations effective doubles.

The work of testing these combinations currently falls the maintainers of which there is current only one, me, and i am rather buys with other things.

The all-or-norhing approach to the external libs was chosen a long time ago to reduce my work load. I have less time to work on libsndfile now than I did back then, so i am very much against this change.

arthurt:
I appreciate that being the sole maintainer of this project is a lot of work, and thank you for it. And certainly increasing your workload is something I would want to avoid. However, unless I'm missing something (always possible) aren't the external codecs are pretty much orthogonal to each other? And if so their testable interactions with each other nonexistant?

Each codec only calls its external libraries functions in one file, and these files only have one non-static funciton, blah_open(), which is always defined and compiles to a stub if support is missing. If two external libraries collide somehow, the collision would be present if testing with all external libraries, and not requiring testing all 2^n possibilities.

Ogg is slightly different, as ogg-embedded codecs need Ogg. However, Ogg does not depend on the existence of particular codec. It interacts with them through their ogg_*_open() functions only, which are of course stubs if they are missing. Ogg dependent codecs* of course compile to stubs if Ogg isn't present.

One thing testing all 2^n possibilities could do is test for if support for one codec erroneously requires support for another which is not present. But, only one codec can be in use at a time, and again, the only intractable part is function which will be an error-returning stub.

Another thing testing all 2^n possibilities could do is reveal broken library dependencies, but that is bug in the external libraries and their pkgconfig/config-program/documentation, not libsndfile.

The other behaviour in libsndfile which is selectively triggered by the existence of external libraries are in the results of sf_command() on SFC_GET_(SIMPLE_)?FORMAT.* and tests. Again, the presence or non-presence here remain orthogonal to each other.

*FLAC is not an Ogg dependent codec. While Flac may be embedded in Ogg, libsndfile doesn't deal with this, and ogg_open() throws the file to libFLAC to do the Ogg'ing. Really, sf_open should be patched so 'OggS' headed files are tried for FLAC if Ogg support isn't present.

erikd:
> And if so their testable interactions with each other nonexistant?

Sure, if the code compiles then all is fine. The problem is that to may these optional requires CPP `#ifdef` jiggerly pokery which is *highly* fragile and impossible to test without at least compiling all combinations.

evpobr:
Vote for this. On Linux is it easy to get all required packages, but on Windows is it not. Flexibility is good.

> ... and impossible to test without at least compiling all combinations.

I think if test with all external libraries enabled is passed, it is very unlikely that test with some codec disabled will fail. So it is just wasting of time.

erikd:
> I think if test with all external libraries enabled is passed, it is very unlikely that test with some codec disabled will fail. So it is just wasting of time.

I think you are wrong. I can think of at least one situation which could go wrong.

Say we have two libraries `FOO` and `BAR` and code that does this:

```
#if HAVE_LIB_FOO

// 100+ lines of code

#if HAVE_LIB_BAR

// 100+ lines of code

#endif

// 100+ lines of code

#endif
```
If both `HAVE_LIB_FOO` and `HAVE_LIB_BAR` are enabled all is good. If neither `HAVE_LIB_FOO` nor `HAVE_LIB_BAR` are enabled all is good.

However some combination of having one defined and the other not is highly likely to result in a compile error and as the number of combinations increase, the likelihood of something going wrong also increases.

evpobr:
Ok, then close it is time to close issue.

arthurt:
Yes, CPP selectable code like that can be fragile. However, I argue that in this case it is not. The use of CPP defines for the codecs is much simpler. With the current architecture of libsndfile, the hypothetical condition of two nested HAVE_codec CPP if/else 100 lines apart would not usefully occur in code.

Codec source files are all of the type:

<copyrights, initial headers>
#if HAVE_FOO
<actual code>
#else 
<stub>
#endif 

Which is only slightly more complex than a header guard.

The codecs do not reference each other, so do not use more than one CPP if/else statement. That if/else may check multiple conditions, eg (HAVE_EXPERIMENTAL && HAVE_SPEEX) or (HAVE_OGG && HAVE_VORBIS) but it only occurs once for the whole file, at the top.

Inside command.c there are multiple checks for HAVE_FLAC or (HAVE_OGG && HAVE_VORBIS) but they select whether a codec is reported through the sf_command interface, never more than three lines, usually one, and where there is no logic.

Likewise, tests are never more than

#if HAVE_foo
<test>
#else 
<report test skipped>
#endif

erikd:
> the hypothetical condition of two nested HAVE_codec CPP if/else 100 lines apart would not usefully occur in code.

Actually the occurrence of two `#ifdef`s like this *would* happen in the code for FLAC if someone added support for FLAC in an Ogg container.  Whether they would be nested is another matter.

CPP is horribly fragile and when it gets inadvertently broken, it can be months of even years before its discovered.

arthurt:
> Actually the occurrence of two #ifdefs like this would happen in the code for FLAC if someone added support for FLAC in an Ogg container. Whether they would be nested is another matter.

I would suggest a better solution would be to put ogg-in-flac support in its own file: ogg_flac.c. Similarily, if Opus in something other than Ogg was to be supported, I think it would make sense to leave ogg_opus.c alone and implement it in a different source file.

The crux of my argument is that using CPP to selectively enable or disable codecs is safe /if/ using the current architecture of only using them once over the whole file. If functionality requires nesting CPP checks for codecs within a source file, something is wrong. I propose that as a rule rather than all-or-nothing dependencies.

PR #405, opened 2018-08-19
CMake: updates & fixes

### LINKED ISSUE #289: Windows msvc cmake pkg-config: tiny bug in libname generation
Hi,

the sndfile.pc generated in the msvc/cmake build gives sndfile as name for the library, while the real name is libsndfile-1. This causes pkg-config lookups to produce wrong data and linking to fail. Admittedly using pkg-config in msvc builds is not exactly common, but I managed to bump into a case, trying to build aubio. Aubio uses the waf buildsystem and depends on pkg-config to detect libsndfile.

As we are at it: the build option `ENABLE_PACKAGE_CONFIG` seems to have no effect. My understanding is that it controls if the sndfile.pc file is generated or not (I might misunderstand though). Fortunately the file is always generated, so no prob ;)

evpobr:
# Summary of changes

* Let CMake select library name, it is platform dependent. On MinGW it is `libsndfile-1.dll`, on Cygwin - `cygsndfile-1.dll`, on Win32 - `sndfile.dll`. `pkg-config` knows platform rules and will resolve all these names, closes #289.
* Use only one library target, `sndfile`. No separate static\shared targets to avoid name clashes. Testing disabled for shared library build, it depends on functions, not exported by shared library.
* Impove package config find module. All targets are put in `SndFile::` namespace; library target renamed to `sndfile`. To link library use:

      find_package (SndFile::sndfile REQUIRED)
      target_link_library(helloworld PRIVATE SndFile::sndfile)

* Speed up configuration. Don't check unused functions\headers\features etc.
* Improve external libraries search. Display found library path and version (if detected), make and link imported libraries.
* Rename `DISABLE_EXTERNAL_LIBS` option name to `ENABLE_EXTERNAL_LIBS` for consistance.
* Rename `DISABLE_CPU_CLIP` option name to `ENABLE_CPU_CLIP` for consistance.
* Hide options not relevant for current configuration with `cmake_dependent_option`.
* Move GNU C warnings flags out of CMake script. It is not requirement. Applied to Travi CI builds as as command line parameter `CMAKE_C_FLAGS`.

To @erikd , now i guess  CMake support is close to stable.

To @bagong: check it please.


Nelson-numerical-software:
Hi,
My two cents, "on Win32 - sndfile.dll", why do you need to change the name ? previously it was libsndfile-1.dll.
It will require to all external projects using libsndfile as dynamic library to modify some dependencies.
Thanks

evpobr:
@Nelson-numerical-software , i don't change it now, CMake selects proper name for every platform.

The main advantage of the new naming scheme is that the library linked on every platform by the same name - `sndfile`.

`libsndfile-1.dll` is MinGW build name, and because it was the only way to build `libsndfile` DLL - with MinGW + Autotools, this name became popular outside the platform.

I want to make things right.

evpobr:
New `ENABLE_COMPATIBLE_LIBSNDFILE_NAME` option is added, ON by default for Win32 DLL configuration.

When ON, the DLL name is `libsndfile-1.dll` (for compatibility), when OFF - `sndfile.dll` (as it should be on native Win32 platform).

Static library name on Win32 platform is the same - `sndfile.lib`.

DLL Import library name IS CHANGED, he name is selected by CMake and now it is `sndfile.lib` instead of `libsndfile-1.lib`, we need it to to fix #289.

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ if (NOT ENABLE_CPU_CLIP)
 	set (CPU_CLIPS_POSITIVE FALSE)
 	set (CPU_CLIPS_NEGATIVE FALSE)
 endif ()
+cmake_dependent_option (ENABLE_COMPATIBLE_LIBSNDFILE_NAME "Set DLL name to libsndfile-1.dll (canonical name), sndfile.dll otherwise" ON "WIN32;NOT MINGW;BUILD_SHARED_LIBS" OFF)
```
erikd: Should that not be the default? The default should be no change.

erikd:
Thanks!

bagong:
Just as a anecdotal addendum, not to make a point or create an issue: as the name of the library is libsndfile, the dynamic library should be named liblibsndfile in unixy flavors (including mingw-Windows, but not VS Windows). Some projects (including supercollider) do actually assume the name of the library is sndfile ;) Thanks for all the work!

PR #406, opened 2018-08-19
Opus Support

arthurt:
For consideration and review, Opus codec support. ~~This is currently experimental, has no tests, and only builds when experimental support is enabled.~~

Supports encoding and decoding, metadata (tags), bisection seek support.

Additional functionality was added to the Ogg container code, making it available for other codecs, although using such functionality in these codecs is not a part of this pull request. Introduces ogg_vcomment for parsing and writing Vorbis Comment style tags. This is the tag format used by most Ogg encapsulated streams; OggVorbis, Speex, Opus, OggPCM and OggFLAC. Currently only consumed by Opus.

Introduces SF_CODEC_OPUS subtype, ~~but hides support for it unless experimental code is enabled.~~


erikd:
Wow! Its all green!

I will take a proper look at this over the weekend.

arthurt:
Any feedback so far? I know it's rather a large pull.

erikd:
Looks good so far. Needs tests though :). Was going to add them myself, but life and work have been way too hectic.

arthurt:
An update of where this is at. I've added tests for Opus, basically the same as for Vorbis:
- floating_point_test
- external_libs_test
- format_check_test
- ogg_opus_test (similar to ogg_test)
- compression_size_test
- lossy_comp_test
- string_test

Currently the build passes, but that tells us nothing as Opus support is not enabled by default. I have a branch, https://github.com/arthurt/libsndfile/tree/ogg-opus-testing, in which Opus is enabled by default. The tests pass on both Linux targets in Travis, but fail identically for the Apple target and both x86 and x64 targets of AppVeyor. Currently investigating.

erikd:
@arthurt Looks good so far. Please feel free to tie Opus support in with the FLAC and Vorbis support (ie Opus is supported when the others are supported) with this patch set.

arthurt:
Still having issues with the lossy_comp_test, but I know the reason now. The problem is the libopus version, not the platform. In short, Opus 1.3 passes with a margin of 0.57-0.55, while Opus 1.2 an earlier requires a margin that is much higher at 1.16. The part of the test which fails is a seek then comparison between the output sample and the original data at the same offset.

I've verified that seeking then decoding generates exactly the same output as decoding from the beginning to the seek point. The problem appears to be an excess of amplitude in the Opus 1.2 output, which is okay for the smoothed sample comparison of sdlcomp_test_* , but fails with the seek then exact (non-smoothed) sample comparison.

I think Opus 1.3 has been picked up by 'vcpkg' and so should be in use by Appveyor soon, but really it's all quite foreign to me. This pull still sets the minimum required Opus version to 1.1, as that is the minimum API requirement, but Opus 1.3 is required to pass the test suite.

erikd:
If Opus 1.3 is required to make the tests pass, then I don't mind making 1.3 a requirement.

arthurt:
I would seem that Opus 1.3 is required to pass lossy_comp_test without ridiculous bounds, but only for Windows (32/64) and Mac OS X, Opus 1.1.2 and up compiled with GCC on Linux works just fine. Not sure why, but at least it is consistent from 1.3 on. 1.1 is the minimum API required.

Setting a requirement of 1.3 would fix things, but sadly there is currently no version checking of external libraries in the CMake files for platforms that don't use pkg-config, so that would have to be written.

In the meantime I have written a run-time check in lossy_comp_test that uses the version of libopus as reported in the libsndfile error log to selectively skip the test for older version. Annoyingly this always skips on Windows, as the libopus as built and installed by `vcpkg` reports its version string as 'unknown.'

erikd:
@evpobr Is detecting Opus version in CMake via `pkg-config` something you would be able to help with?

evpobr:
> Setting a requirement of 1.3 would fix things, but sadly there is currently no version checking of external libraries in the CMake files for platforms that don't use pkg-config, so that would have to be written.

No idea how to do that. Common fix is to parse library header with version defines. But `opus` has no such defines.

Another wat is to create package config module when installing. This is perfect solution, you will get all the info you need including version variable, but `opus` port lacks of this feature. It is interesting can i add it for Visual Studio based port, i will take a look.

> I have written a run-time check in lossy_comp_test that uses the version of libopus as reported in the libsndfile error log to selectively skip the test for older version. Annoyingly this always skips on Windows, as the libopus as built and installed by `vcpkg` reports its version string as 'unknown.

Probably is it `vcpkg` port bug. `opus` uses script to generate private version header, may be is fails to work.

evpobr:
> Annoyingly this always skips on Windows, as the libopus as built and installed by `vcpkg` reports its version string as 'unknown.'

Opus Vcpkg port uses bundled Visual Studio solution, which generates `version.h` with `genversion.bat` script. But it fails, so `PACKAGE_VERSION` is set to `unknown`. I don't know why it happens.

arthurt:
Bumping the Opus requirement to 1.3 removes the need to check for it in the test suite. However the test suite may still fail for Windows if consuming Opus through `vcpkg`. This is because `vcpkg`'s ports entry for Opus doesn't interact with how Opus expects it to be built on Windows, and the resulting Opus headers and library lack version numbers. The reason that Opus support still builds on Windows with `vcpkg` is that FindOpus.cmake, failing to use `pkg-config` accepts any library named Opus it can find.

The question is, should this be fixed? Doing so will mean Opus through `vcpkg` will not work until `vcpkg`'s Opus port is fixed (or Opus does versioning some other way.) As this pull makes Opus, Ogg and Flac all-or-nothing, this will break Opus, Ogg and Flag for Windows platforms using `vcpkg` (like AppVeyour.)

(I have no idea what happens when you use Opus through `vcpkg` for Linux...)

Not checking for a dependencies version information in certain cases feels like the wrong thing to do. However, nobody has complained thus far, and `vcpkg`'s versioning of ports is, well, just 'the latest', whatever that might be. The current `vcpkg` repository builds Opus 1.3, so the test suite will pass with it.

However, maybe forcing this to fail correctly will be the squeak for this particular wheel of `vcpkg`'s ports to get the grease it needs.

Thoughts?

evpobr:
As you mentioned above, Appveyor test is correct because Vcpkg has Opus 1.3. So we have everything tested well.

I have idea how to fix that Opus bug. Wrong path in custom command to run `genversion.bat` which generates `version.h`. But it will take time to fix it in `opus` and `vcpkg` repos. Should we wait? I guess not.

I think it is reasonable to terminate `lossy_comp_test` test with without error with warning message if all conditions are satisfied:

* Test reports wrong results (as if Opus version < 1.3)
* `_WIN32` is defined
* Opus reported version as "unknown"

arthurt:
> I have idea how to fix that Opus bug.
Awesome!

Something I forgot in my last comment is bumping Opus to 1.3 will also prevent Opus (and therefore Ogg, Flac, and Vorbis) from building in Travis on Linux as the Travis environment is Ubuntu Xenial, which only has Opus 1.1.2.

arthurt:
> Rename opus::opus target to Opus::opus.

Okay. Just checking but why not `Opus::Opus`?

arthurt:
> > > Rename opus::opus target to Opus::opus.
> > 
> > 
> > Okay. Just checking but why not `Opus::Opus`?
> 
> No problem. There is no clear convension.

Okay. Changed. I chose Opus::Opus as it seemed to match Vorbis::Vorbis, etc.

fredvs:
Fantastic!

And many thanks to boost the Opus project.

Fre;D

arthurt:
@erikd Thoughts on merging this?

Version checking is somewhat stuck on external projects. The runtime test for the tests works for now, can be removed once vpkg is fixed.

erikd:
Any chance you could rebase against master and clean up the commit history. Github can squash commits but afaiac it makes a mess of it.

arthurt:
Okay, rebased. Let me know if there's anything else!

fredvs:
EXCELLENT!

By the way, are you able to compile libsndfile on Windows with Clang (I fail) ?
If you can, uploading the Windows-binary will be highly appreciated. 
But uploading the Windows-binary compiled by gcc.exe or Visual Studio will be nice too.

This only for Windows, compiling libsndfile for Unix (Linux or FreeBSD) is out of the box.

Many thanks for your great work.

Fre;D

evpobr:
> If you can, uploading the Windows-binary will be highly appreciated.

@erikd [compiles](http://www.mega-nerd.com/libsndfile/#Download) binaries each release.

So the only thing we need is new release.

I will make experimental Visual Studio 2017 build and attach to GitHub releases page.

fredvs:
> @erikd [compiles](http://www.mega-nerd.com/libsndfile/#Download) binaries each release.
> So the only thing we need is new release.

Nice ;)
Thanks.

Fre;D

erikd:
Thanks @arthurt .

We will still need to do extensive fuzz testing of these before a release with Opus support.

PR #407, opened 2018-08-25
modernize cmake module include.

conradjones:
using the existing method of the list CMAKE_MODULE_PATH passed to include was causing multiple paths to be appended to one huge invalid path.

see docs here: CMake does this for you, with one caveat as long as there isn't a built in module with the same name, 

https://cmake.org/cmake/help/latest/command/include.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ list (APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 # Setup definitions
 #
 
-include(${CMAKE_MODULE_PATH}/SndFileChecks.cmake)
+include(SndFileChecks)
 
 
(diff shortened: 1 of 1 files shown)


conradjones:
just a heads up, looks like there are further issues with the 3 CMake
related commits from the 25th, it still fails to build via conan. checking
out the previous commit from the 10th builds find via conan.

I'll try and get a sample project together and open an issue to demonstrate
it, but I''m not sure when I'll get time to do it

conradjones:
Just to follow up on this I found it wasn't an issue with the new CMake
related commits, you've done a great job on them it was due FindXXX.cmake
files exported by my conan packages for the dependencies (ogg/flac etc) and
I stopped exporting them in my conan package and let it use your scripts in
libsndfile to find the libraries and it worked fine finding them in the
conan package directories

PR #409, opened 2018-08-30
README.md: cleanup and explain more about CMake

evpobr:


README.md
```diff
@@ -13,7 +13,7 @@ The canonical source code repository for libsndfile is at
 
 You can grab the source code using:
 
-    $ git clone git://github.com/erikd/libsndfile.git
+    \$ git clone git://github.com/erikd/libsndfile.git
```
erikd: In the Github view of this markdown, the backslash remains.

README.md
```diff
@@ -22,27 +22,26 @@ a more limited and experimental CMake based build system. Use of the CMake build
 system is documented below.
 
 Setting up a build environment for libsndfile on Debian or Ubuntu is as simple as:
-```
-sudo apt install autoconf autogen automake build-essential libasound2-dev \
+
+    sudo apt install autoconf autogen automake build-essential libasound2-dev \
     libflac-dev libogg-dev libtool libvorbis-dev pkg-config python
-````
+
```
erikd: I actually *like* the fact that the second line is indented. In programming languages like Python and Haskell an indented second line means that it is semantically connected to the first.

evpobr: No problem, i'll fix this.

README.md
evpobr: Funny, markdownlint generates [warning](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md014---dollar-signs-used-before-commands-without-showing-output). Do we need $ here?

erikd: Yes, i just `$` without the backslash is correct,

evpobr: I mean we can remove `$` here safely. Comment from [MarkDown Style Guide](http://www.cirosantilli.com/markdown-style-guide/):

> *Don’t* prefix shell code with dollar signs $ unless you will be showing the command output on the same code block.
>
> If the goal is to clarify what the language is, do it on the preceding paragraph.
>
> Rationale: harder to copy paste, noisier to read.
>
> Good:
>
>     echo a
>     echo a > file
> Bad:
>
>     $ echo a
>     $ echo a > file
> Good, shows output:
>
>     $ echo a
>     a
>     $ echo a > file
> Good, language specified on preceding paragraph:
>
> Use the following Bash code:
>
>     echo a
>     echo a > file

evpobr: I'll check other places.

erikd:
Thanks!

PR #426, opened 2018-10-09
Fix lossy_comp_test sdlcomp_test_(float|double) for Vorbis

arthurt:
While experimenting, determined that Vobis' analysis stops working on
non-normalized floating point input outside the range of [2000.0, -2000.0], and
seems to get worse the larger the magnitude. As the sdlcomp_test_float and
sldcomp_test_double tests work with non-normalized floats with a magnitude of
32000.0, this explains why they do not work for Vorbis.

This patch fixes the test by reducing the magnitude of the test data at encode
and restoring it at decode if it detects Vorbis, mantaining the same 32000.0
overall magnitude so the envelope values for the test are the same as they are
for other tests and format.

It is unknown at this time if the input range magnitude of 2000.0 is  a
limitation of Vorbis or libvorbis.

### DIFF
--- a/tests/lossy_comp_test.c
+++ b/tests/lossy_comp_test.c
@@ -1821,18 +1821,21 @@ sdlcomp_test_float	(const char *filename, int filetype, int channels, double mar
 	int				k, m, seekpos ;
 	sf_count_t		datalen ;
 	float			*orig, *data, *smooth ;
-	double			half_max_abs ;
+	double			half_max_abs , scale ;
 
 channels = 1 ;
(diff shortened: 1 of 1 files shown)


arthurt:
This all came out of working on tests for the ogg-opus branch. Adding Opus tests to lossy_comp_tests will require a similar fix as libopus only allows normalized floating point values, possibly depending on whether libopus was compiled with or without floating point support.

It seemed to make sense to me to offer this as a stand-alone pull request.

PR #438, opened 2018-12-31
Handle snprintf() format warnings

### LINKED ISSUE #437: src/mpc2k.c compilation fails with 'error: ‘%s’ directive output may be truncated writing up to 255 bytes into a region of size 18 [-Werror=format-truncation=]'
On Ubuntu 18.10 (gcc 8.2.0), when trying to compile master (8ddc442) using the README.md recommended workflow

```sh
./autogen.sh
./configure --enable-werror
make
```

it fails with

```gcc
[…]
  CC       src/libsndfile_la-mpc2k.lo
src/mpc2k.c: In function ‘mpc2k_write_header.part.0’:
src/mpc2k.c:134:48: error: ‘%s’ directive output may be truncated writing up to 255 bytes into a region of size 18 [-Werror=format-truncation=]
  snprintf (sample_name, sizeof (sample_name), "%s                    ", psf->file.name.c) ;
                                                ^~
In file included from /usr/include/stdio.h:873,
                 from src/mpc2k.c:21:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:67:10: note: ‘__builtin___snprintf_chk’ output between 21 and 276 bytes into a destination of size 18
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

It seems that gcc has been getting too picky for that space-fill hack. Do you prefer disabling `-Werror` or would you prefer a pull request that space-fills differently?

MarcelWaldvogel:
Fixes #437 (and several other similar warnings treated as errors)

### DIFF
--- a/src/chunk.c
+++ b/src/chunk.c
@@ -219,6 +219,7 @@ psf_save_write_chunk (WRITE_CHUNKS * pchk, const SF_CHUNK_INFO * chunk_info)
 {	union
 	{	uint32_t marker ;
 		char str [5] ;
+		/* Update snprintf() format string below when changing this */
 	} u ;
 	uint32_t len ;
 
@@ -241,7 +242,7 @@ psf_save_write_chunk (WRITE_CHUNKS * pchk, const SF_CHUNK_INFO * chunk_info)
 	len = chunk_info->datalen ;
 	while (len & 3) len ++ ;
 
-	snprintf (u.str, sizeof (u.str), "%s", chunk_info->id) ;
+	snprintf (u.str, sizeof (u.str), "%.4s", chunk_info->id) ;
 
 	pchk->chunks [pchk->used].hash = strlen (chunk_info->id) > 4 ? hash_of_str (chunk_info->id) : u.marker ;
 	pchk->chunks [pchk->used].mark32 = u.marker ;
--- a/src/mpc2k.c
+++ b/src/mpc2k.c
@@ -131,7 +131,7 @@ mpc2k_write_header (SF_PRIVATE *psf, int calc_length)
 	if (psf->is_pipe == SF_FALSE)
 		psf_fseek (psf, 0, SEEK_SET) ;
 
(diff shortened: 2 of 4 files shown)


src/chunk.c
```diff
@@ -241,7 +242,7 @@ psf_save_write_chunk (WRITE_CHUNKS * pchk, const SF_CHUNK_INFO * chunk_info)
 	len = chunk_info->datalen ;
 	while (len & 3) len ++ ;
 
-	snprintf (u.str, sizeof (u.str), "%s", chunk_info->id) ;
+	snprintf (u.str, sizeof (u.str), CHUNK_FORMAT_STR, chunk_info->id) ;
```
erikd: Is it not just possible to use `"%.4s"` there instead of `CHUNK_FORMAT_STR` ?

src/mpc2k.c
```diff
@@ -131,7 +132,7 @@ mpc2k_write_header (SF_PRIVATE *psf, int calc_length)
 	if (psf->is_pipe == SF_FALSE)
 		psf_fseek (psf, 0, SEEK_SET) ;
 
-	snprintf (sample_name, sizeof (sample_name), "%s                    ", psf->file.name.c) ;
+	snprintf (sample_name, sizeof (sample_name), HEADER_FORMAT_STR, psf->file.name.c) ;
```
erikd: Again, why a macro there instead of the format string itself?

erikd:
I think I prefer regular format strings rather than these macros. Comments on  the format strings may help though.

erikd:
Please squash these commits into one and rebase the branch against master.

erikd:
Thanks!

Issue #277, opened 2017-05-31
fails with multipart Ogg/Vorbis files


umlaeute:
(this was reported as https://bugs.debian.org/#761240)

Ogg/Vorbis files can consist of multiple parts, simply concatenated
together:

```sh
$ head -c 10000 /dev/zero | oggenc -r - -o foo.ogg
$ head -c 10000 /dev/zero | oggenc -r - -o bar.ogg
$ cat foo.ogg bar.ogg > baz.ogg
```

But libsndfile doesn't handle those files correctly. While it shows
the correct size:

```sh
$ sndfile-info baz.ogg
[...]
Frames      : 5000
Channels    : 2
```

It only reads the frames of the first part (the bug is in the
library, not in sndfile-convert; I get the same result with my own
programs using the library):

```sh
$ sndfile-convert -pcm16 baz.ogg baz.wav
$ sndfile-info baz.wav
[...]
Frames      : 2500
Channels    : 2
```

To fix it, when using libvorbisfile, one would just check the result
of `ov_read_*` for `OV_HOLE` and retry, but libsndfile doesn't use
libvorbisfile, but a lower-level interface ...

arthurt:
Since #457 libsndfile should at least be consistent and only report the size of the first part now.

PR #457, opened 2019-02-21
Vorbis: Use new ogg functions, bisection seek support.

arthurt:
Use new Ogg functionality merged with Opus support to reduce code duplication, simplify Vorbis support (decoding in particular) and improve seek performance.

This uses the bisection-based seek support, rather than decode-from-start seeking. This removes the need for a separate file decode for calculating file sample count. The 'nasty nest' is simplified and flattened out a bit.

No new tests are introduced, as no new functionality is exposed (other than less computationally expensive seeking.)

Had this on the back-burner ready to go once the Ogg changes required by the Opus pull request were merged.

### DIFF
--- a/src/ogg_vorbis.c
+++ b/src/ogg_vorbis.c
@@ -231,11 +231,11 @@ vorbis_read_header (SF_PRIVATE *psf)
 	psf->dataoffset	= ogg_sync_ftell (psf) ;
 
 	/*
-	** Caculate the granule position offset. The first page with a payload
-	** packet shouldn't end in a continued packet. The difference between the
-	** page's granule position and the sum of frames on the page tells us the
-	** granule position offset.
-	** See https://xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-132000A.2
+	**	Caculate the granule position offset. The first page with a payload
+	**	packet shouldn't end in a continued packet. The difference between the
+	**	page's granule position and the sum of frames on the page tells us the
+	**	granule position offset.
+	**	See https://xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-132000A.2
 	*/
 	ogg_stream_unpack_page (psf, odata) ;
 	vdata->pcm_start = odata->pkt [odata->pkt_len - 1].granulepos ;
@@ -246,10 +246,13 @@ vorbis_read_header (SF_PRIVATE *psf)
 	else
 		vdata->pcm_start = 0 ;
 
+	/*
+	**	Find the end of the stream, save it. Only works if the file is seekable.
+	*/
 	vdata->loc = vdata->pcm_start ;
 	vdata->pcm_end = (uint64_t) -1 ;
 	psf->datalength = psf->filelength ;
-	if (psf->sf.seekable)
+	if (!psf->is_pipe)
 	{	sf_count_t last_page ;
 		sf_count_t saved_offset ;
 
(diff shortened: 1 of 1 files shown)


erikd:
Nicely done. Thanks!

fredvs:
@arthurt: Excellent, it works like charm.

Huh, maybe you may show your work to Opus-project.
(Like explained before, with your code, no problems applying DSP from some external libraries).

Congrats for your perfect code.

Fre;D

fredvs:
@arthurt:

Now tat you are deep inside, is it possible to do Opus internet URL streaming too with libsndfile?

With libopusfile, I use those methods to do create the working web-streaming handle:

``op_test_callbacks(), op_test_open(), op_link_count().``

Would it be possible to do this with libsndfile?

And what about the Opus feature OPUS_APPLICATION_AUDIO/VOIP and friends (maybe only _AUDIO for libsndfile)?

Thanks.

Fre;D

arthurt:
> Huh, maybe you may show your work to Opus-project.

Most of the business parts of the code merged comes from Xiph.org sources such as libopus and libopusfile. The work of this pull request was shuffling it around so it fit better to libsndfile.

> Now that you are deep inside, is it possible to do Opus internet URL streaming too with libsndfile?
> 
> With libopusfile, I use those methods to do create the working web-streaming handle:
> 
> `op_test_callbacks(), op_test_open(), op_link_count().`
> 
> Would it be possible to do this with libsndfile?

I think that is outside of the scope of libsndfile to implement that directly (not my call though, I'm just a volunteer.) However, it is possible to get libsndfile to read from sockets, either using `sf_open_fd` or `sf_open_virtual` and writing your own appropriate IO methods.

For the `op_link_count()`, this is currently unsupported by the code merged. Like Ogg files, multiple Opus files concatenated together makes one legal Opus file. In Ogg parlance, the logical streams are said to be chained, and each logical stream (what could be a stand-alone file) is called a link. This is not currently supported by libsndfile, and only the first link of a Ogg-based file can be read, and only one link may be written. It might still be possible to manipulate such chained Ogg/Opus files using libsndfile's embedded file mechanisms. See issue #277.

> And what about the Opus feature OPUS_APPLICATION_AUDIO/VOIP and friends (maybe only _AUDIO for libsndfile)?

When initializing an encoder you must chose one of the AUDIO or VOIP application settings. Currently this is hardcoded to AUDIO and not select-able by the end user of libsndfile as this is probably the choice that is always wanted for libsndfile's use cases. Choosing the VOIP profile reduces the encoders latency and packet sizes a bit, and is perfect for feeding an RTP stream. The trade-off is many small packets which will add overhead when put into an Ogg stream, and possibly worse compression from less analysis (lower latency)

It may be possible that choosing the VOIP setting better tunes the encoder for speech content, so maybe that is a reason to expose it, but I should expect that Opus probably does the right thing with speech anyways when using AUDIO.

fredvs:
Hello Arthur.

Thanks for your work and your clear answers.

> Most of the business parts of the code merged comes from Xiph.org sources such as libopus and libopusfile.

Ha, so very strange that your code + libsndfile is working with some external libraries (see previous post) and libopusfile.so crash using it.  And this crash appends also with audacity (that uses also libopusfile.so).  Anyway libsnfile will fix ll this.

 > However, it is possible to get libsndfile to read from sockets, either using `sf_open_fd` or `sf_open_virtual` and writing your own appropriate IO methods.
>  For the `op_link_count()`, this is currently unsupported by the code merged.

Ha, ok, I will try this and maybe a workaround for `op_link_count() ` can be found.

> When initializing an encoder you must chose one of the AUDIO or VOIP application settings. Currently this is hardcoded to AUDIO and not select-able

Alright, IMHO voip should be mainly useful for web-streaming.
Maybe it is better that I still use for the moment  libopusfile for web-streaming voip (voice over ip).
(Even if I would prefer a unique libsndfile for all audio stuffs).

Thank for your heroes work Arthur.

Fre;D

PR #462, opened 2019-03-27
CMake: Fix installation of man pages.

mlehtima:
Incorrect define check and non-existent argument used for install.


evpobr:
Thanks @mlehtima

PR #463, opened 2019-04-01
Fixes to Opus handling

erikd:
Fix a small number of errors found by fuzz testing. Also some minor stylistic fixes.

@arthurt  Would really appreciate your review of these changes. If you need me to, I can split these into fuzzing fixes and cosmetic change commits.

### DIFF
--- a/programs/common.c
+++ b/programs/common.c
@@ -1,5 +1,5 @@
 /*
-** Copyright (C) 1999-2017 Erik de Castro Lopo <erikd@mega-nerd.com>
+** Copyright (C) 1999-2019 Erik de Castro Lopo <erikd@mega-nerd.com>
 ** Copyright (C) 2008 George Blood Audio
 **
 ** All rights reserved.
@@ -282,6 +282,7 @@ typedef struct
 	int			format ;
 } OUTPUT_FORMAT_MAP ;
 
+/* Map a file name extension to a container format. */
 static OUTPUT_FORMAT_MAP format_map [] =
 {
 	{	"wav", 		0,	SF_FORMAT_WAV	},
@@ -316,6 +317,7 @@ static OUTPUT_FORMAT_MAP format_map [] =
 	{	"prc",		0,	SF_FORMAT_WVE	},
 	{	"ogg",		0,	SF_FORMAT_OGG	},
 	{	"oga",		0,	SF_FORMAT_OGG	},
+	{	"opus",		0,	SF_FORMAT_OGG	}, /*  Opus data in an Ogg container. */
 	{	"mpc",		0,	SF_FORMAT_MPC2K	},
 	{	"rf64",		0,	SF_FORMAT_RF64	},
 } ; /* format_map */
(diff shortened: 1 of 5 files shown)


src/ogg_opus.c
```diff
@@ -1043,6 +1046,17 @@ ogg_opus_read_refill (SF_PRIVATE *psf, OGG_PRIVATE *odata, OPUS_PRIVATE *oopus)
 			} ;
 		} ;
 
+	if (oopus->len > oopus->buffersize)
```
arthurt: Did this come out from fuzzing? If `oopus->len > oopus->buffersize` here an error has occurred.

Either opus_decode_... decoded more samples than the buffer can hold (which should never happen as it should return OPUS_BUFFER_TOO_SMALL and fail.)

Or, the end-of-stream padding cut off code above was fed bad data. An alternative would be to alter line 1031:
```
oopus->len = SF_MIN (nsamp, SF_MAX (0, (int) (oopus->pg_pos - oopus->pkt_pos) / oopus->sr_factor)) ;
```
That is, bound `len` to [0, nsamp].

arthurt: Or if `oopus->len > nsamp`, bail out with SFE_MALFORMED_FILE;

programs/common.c
```diff
@@ -316,6 +316,8 @@ static OUTPUT_FORMAT_MAP format_map [] =
 	{	"prc",		0,	SF_FORMAT_WVE	},
 	{	"ogg",		0,	SF_FORMAT_OGG	},
 	{	"oga",		0,	SF_FORMAT_OGG	},
+	{	"opus",		0,	SF_FORMAT_OGG	},
+	{	"opus",		0,	SF_FORMAT_OPUS	},
```
arthurt: I don't think this could ever be hit, as the `"opus"` above would mask it.

Also, not sure how to accomplish this, but mapping "opus" to SF_FORMAT_OGG or SF_FORMAT_OPUS alone seems wrong. Could it be possible to change this to map "opus" to SF_FORMAT_OGG | SF_FORMAT_OPUS, but then one is mapping a subtype as well.

`sfe_file_type_of_ext` could be modified to detect if format_map[].format has set a subtype, and if so ignore the format passed as an argument. Thoughts?

src/ogg_opus.c
erikd: Yeah, that was something found by fuzzing.

programs/common.c
erikd: Yeah, you're right. That was some last second futzing before I put the computer down for the evening.

programs/common.c
```diff
@@ -423,6 +425,7 @@ sfe_container_name (int format)
 		case SF_FORMAT_OGG : return "OGG" ;
 		case SF_FORMAT_MPC2K : return "MPC2K" ;
 		case SF_FORMAT_RF64 : return "RF64" ;
+		case SF_FORMAT_OPUS : return "OPUS" ;
```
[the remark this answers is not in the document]
erikd: ? RF64 and OPUS are both subtypes.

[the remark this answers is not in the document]
erikd: Yep that should not be there.

src/ogg_opus.c
arthurt: Humm. It's just, I cannot see how reallocating would help here. The effect would be that instead of decoded samples, whatever uninitialized data malloc() returns is output. I think returning SFE_MALFORMED_FILE is a pretty low-risk change in this situation.

erikd: ok, will look at that and the other solution. Then decide if it needs to be fuzzed again.

Issue #78, opened 2014-10-06
Max channels supported is 256


oddeirik:
SF_MAX_CHANNELS defines the maximum number of channels supported at 256. For most formats and use cases I can see that this isn't really an issue. However, WAV and RF64/BWF (and I'm guessing W64) all support up to 65535 channels.

I cloned the source from git and changed SF_MAX_CHANNELS to 65535, compiled the library and successfully created a BWF file with several hundred channels, but I'm not sure how this will affect other parts of the library. As far as I could see, there are several formats with additional channel checks in sf_format_check(), like FLAC.

(If you're wondering why someone would create WAV files with over 256 channels, it's because I'm trying it out for seismic data storage, which generally means lots of channels, although at fairly low sample rates.)

rd--:
hello,

i think i've also run into this problem

the attached next/au file (not written using libsndfile) has 12 frames
and 2000 channels (each channel is the binary encoding of the channel
number)

i can read this file accurately using the haskell libsndfile bindings
(ie. the data as read is equal to the data as written) and
supercollider will load the file (using libsndfile)

but i can't write this data set using the same libsndfile bindings?

sf_format_check fails, and if i leave that out the write fails (with
an UnrecognisedFormat error)

what should be happening here?

i can work around this, since the important part is the supercollider
can read the file, but it seems odd.

best,
rohan

ps. for the use case here (encoding two-dimensional data files,
ie. image files, as audio files to transfer data to the supercollider
synthesiser) 1024 is not quite as good a number as 8192 would be...

rd--:
ps. github won't let me attach a .au file, even base64 encoded with a .txt extension

it's temporarily at http://rd.slavepianos.org/tmp/mc-12-2000.au (96k) in case useful

erikd:
> i can read this file accurately using the haskell libsndfile bindings
> (ie. the data as read is equal to the data as written) 

If libsndfile itself doesn't support this, then how do the haskell bindings do it? I'm familiar with Haskell. Can you provide some example code that reads the file you provided?

rd--:
libsndfile _reads_ these many channel files, but won't write such a file.

ie. the channel count limit doesn't seem to affect reading of (at least) NeXT files

i've added the test files at

  http://rd.slavepianos.org/sw/hsc3-sf/au/

there are equivalent C and haskell printers at

  http://rd.slavepianos.org/sw/hsc3-sf-hsndfile/cmd/sf-text.c
[...]

PR #468, opened 2019-05-28
Increase arbitrary channel limit

nicholasfagan:
Increased SF_MAX_CHANNELS to be 32766 (2^15 - 2). One less than the
largest number that can be supported by a signed 2-byte integer, which
is what I determined is the largest size supported by all formats.

The reason is 2^15 - 2 instead of 2^15 - 1 is because of less-than
comparisons that could probably be less-than-equal-to.

The previous limit of 1024 is arbitrary, and many formats support much
more than that.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -116,7 +116,7 @@
 #define		COMPILE_TIME_ASSERT(e)	(sizeof (struct { int : - !! (e) ; }))
 
 
-#define		SF_MAX_CHANNELS		1024
+#define		SF_MAX_CHANNELS		32766
 
 
 /*


nicholasfagan:
I also noticed issue #78 mentions this topic, and was resolved by commit bc012ab . I hope this topic is not controversial. I require a large number of channels for a project, which includes hundreds of sound channels paired with some non-sound data. Hoping this could be merged so I could use an official release. Thoughts?

erikd:
Supporting large numbers of channels opens up libsndfile to DoS attacks. The increase from 256 to 1024 was really small. Increasing in 32 fold like you suggest is a much bigger risk.

I also *really* wonder what kind of project requires more than 1024 channels.

nicholasfagan:
I'm trying to package lots of sensor data (both audio and other) for later analysis. The current number of sensors does not exceed 1024, but I expect it to in the future. 
Would accept smaller increase?

erikd:
`1024` is a *huge* number of channels. I already think it is larger than anyone could reasonably require for any audio application.

Your application is not a pure audio application I I suggest there are probably file formats for this that libsndfile.I would suggest looking at file formats used in data science.

nicholasfagan:
Ok. Thanks for your time @erikd, I really appreciate the discussion! PR closed.

PR #499, opened 2019-11-20
MPEG Encode/Decode Support

arthurt:
MPEG Encode/Decode Support

Uses libmpg123 for decode, liblame for encode. Encoding and decoding support is independent of each other and is split into separate files. MPEG support is generalized as subformats, SF_FORMAT_MPEG_LAYER(I,II,III) so that it might be used by other containers (MPEG1WAVEFORMAT for example), but also contains a major format SF_FORMAT_MP3 for 'mp3 files.' 

Introduces the command SFC_GET_BITRATE_MODE/SFC_SET_BITRATE_MODE.

Encoding Status
- Layer III encoding
- ID3v1 writing
- ID3v2 writing
- Lame/Xing Tag writing
- Bitrate selection command
- VBR or CBR

Decoding Status
- Layers I/II/III decoding
- ID3v1 reading
- ID3v2 reading
- Seeking

Decoding Todo
- ~~mpg123 global init race~~
- ~~specific error codes~~
- ~~better decoder corruption handling~~

Future Possible Todos
- Add WAV support (WAVE_FORMAT_MPEG1 and WAVE_FORMAT_MPEGLAYER3)
- Add  Layer II encoding by twolame

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -189,6 +189,10 @@ jobs:
         if: startsWith(matrix.os,'ubuntu')
         run: sudo apt-get install -y autogen ninja-build libogg-dev libvorbis-dev libflac-dev libopus-dev libasound2-dev libsqlite3-dev libspeex-dev libmp3lame-dev libmpg123-dev
 
+      - name: Workaround vcpkg yasm-tools for x64-windows
+        if: (startsWith(matrix.os, 'windows') && startsWith(matrix.triplet, 'x64-windows'))
+        run: vcpkg install yasm-tool:x86-windows
+
       - name: Install Windows dependencies
         if: startsWith(matrix.os,'windows')
         run: |
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -28,24 +28,28 @@ jobs:
             os: ubuntu-latest
             cc: gcc
             cxx: g++
+            autotools-options: --enable-werror
             build-system: autotools
 
           - name: ubuntu-clang-autotools
             os: ubuntu-latest
             cc: clang
             cxx: clang++
+            autotools-options: --enable-werror
             build-system: autotools
 
           - name: ubuntu-gcc-ossfuzz
             os: ubuntu-latest
             cc: gcc
             cxx: g++
+            autotools-options: --enable-werror
             build-system: ossfuzz
 
           - name: macos-autotools
             os: macos-latest
             cc: clang
             cxx: clang++
+            autotools-options: --enable-werror
             build-system: autotools
 
           - name: ubuntu-gcc-cmake
@@ -179,16 +183,16 @@ jobs:
       - name: Install MacOS dependencies
         if: startsWith(matrix.os,'macos')
         run: |
-          brew install automake autogen speex
+          brew install automake autogen speex mpg123
 
-      - name: Install Lunux dependencies
+      - name: Install Linux dependencies
         if: startsWith(matrix.os,'ubuntu')
-        run: sudo apt-get install -y autogen ninja-build libogg-dev libvorbis-dev libflac-dev libopus-dev libasound2-dev libsqlite3-dev libspeex-dev
+        run: sudo apt-get install -y autogen ninja-build libogg-dev libvorbis-dev libflac-dev libopus-dev libasound2-dev libsqlite3-dev libspeex-dev libmp3lame-dev libmpg123-dev
 
       - name: Install Windows dependencies
         if: startsWith(matrix.os,'windows')
         run: |
-          vcpkg install libvorbis libflac opus sqlite3 speex --triplet ${{matrix.triplet}}
+          vcpkg install libvorbis libflac opus sqlite3 speex mp3lame mpg123 --triplet ${{matrix.triplet}}
 
       - name: Configure, build and test with Autotools
         env:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,7 @@ include(SndFileChecks)
 
 cmake_dependent_option (BUILD_REGTEST "Build regtest" ON "SQLITE3_FOUND" OFF)
 cmake_dependent_option (ENABLE_EXTERNAL_LIBS "Enable FLAC, Vorbis, and Opus codecs" ON "Vorbis_FOUND;FLAC_FOUND;OPUS_FOUND" OFF)
+cmake_dependent_option (ENABLE_MPEG "Enable MPEG codecs" ON "LAME_FOUND;MPG123_FOUND" OFF)
 cmake_dependent_option (ENABLE_CPU_CLIP "Enable tricky cpu specific clipper" ON "CPU_CLIPS_POSITIVE;CPU_CLIPS_NEGATIVE" OFF)
 if (NOT ENABLE_CPU_CLIP)
 	set (CPU_CLIPS_POSITIVE FALSE)
@@ -104,11 +105,12 @@ set (HAVE_ALSA_ASOUNDLIB_H ${ALSA_FOUND})
 set (HAVE_SNDIO_H ${SNDIO_FOUND})
 
 set (ENABLE_EXPERIMENTAL_CODE ${ENABLE_EXPERIMENTAL})
+set (HAVE_MPEG_LIBS ${ENABLE_MPEG})
 set (HAVE_SPEEX ${ENABLE_EXPERIMENTAL})
-set (HAVE_OPUS ${ENABLE_EXPERIMENTAL})
 
 add_feature_info (BUILD_SHARED_LIBS BUILD_SHARED_LIBS "build shared libraries")
 add_feature_info (ENABLE_EXTERNAL_LIBS ENABLE_EXTERNAL_LIBS "enable FLAC, Vorbis, and Opus codecs")
+add_feature_info (ENABLE_MPEG ENABLE_MPEG "enable MPEG audio (including mp3) codecs")
 add_feature_info (ENABLE_EXPERIMENTAL ENABLE_EXPERIMENTAL "enable experimental code")
 add_feature_info (BUILD_TESTING BUILD_TESTING "build tests")
 add_feature_info (BUILD_REGTEST BUILD_REGTEST "build regtest")
@@ -146,6 +148,18 @@ set_package_properties (FLAC PROPERTIES
 	URL "www.xiph.org/flac/"
 	DESCRIPTION "Free Lossless Audio Codec Library"
 	PURPOSE "Enables FLAC support"
+	)
+set_package_properties (LAME PROPERTIES
+	TYPE OPTIONAL
+	URL "https://lame.sourceforge.io/"
+	DESCRIPTION "High quality MPEG Audio Layer III (MP3) encoder"
+	PURPOSE "Enables MPEG layer III (MP3) writing support"
+	)
+set_package_properties (MPG123 PROPERTIES
+	TYPE OPTIONAL
+	URL "https://www.mpg123.de/"
+	DESCRIPTION "MPEG Audio Layer I/II/III decoder"
+	PURPOSE "Enables MPEG Audio reading support"
 	)
 set_package_properties(Opus PROPERTIES
 	TYPE RECOMMENDED
@@ -247,6 +261,7 @@ add_library (sndfile
 	src/ogg.c
 	src/chanmap.h
 	src/chanmap.c
+	src/id3.h
 	src/id3.c
 	$<$<BOOL:${WIN32}>:src/windows.c>
 	src/sndfile.c
@@ -284,8 +299,12 @@ add_library (sndfile
 	src/ogg_speex.c
 	src/ogg_pcm.c
 	src/ogg_opus.c
+	src/ogg_vcomment.h
 	src/ogg_vcomment.c
 	src/nms_adpcm.c
+	src/mp3.c
+	src/mpeg_decode.c
+	src/mpeg_l3_encode.c
 	src/GSM610/config.h
 	src/GSM610/gsm.h
 	src/GSM610/gsm610_priv.h
@@ -349,6 +368,8 @@ target_link_libraries (sndfile
 		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:FLAC::FLAC>
 		$<$<AND:$<BOOL:${ENABLE_EXPERIMENTAL}>,$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>,$<BOOL:${HAVE_SPEEX}>>:Speex::Speex>
 		$<$<BOOL:${HAVE_EXTERNAL_XIPH_LIBS}>:Opus::opus>
+		$<$<BOOL:${HAVE_MPEG_LIBS}>:MPG123::libmpg123>
+		$<$<BOOL:${HAVE_MPEG_LIBS}>:Lame::Lame>
 	)
 set_target_properties (sndfile PROPERTIES
 	PUBLIC_HEADER "${sndfile_HDRS}"
@@ -1187,6 +1208,14 @@ if (BUILD_TESTING)
 			$<$<BOOL:${LIBM_REQUIRED}>:m>
 		)
 
+	add_executable (mp3_test tests/mp3_test.c)
+	target_link_libraries (mp3_test
+		PRIVATE
+			sndfile
+			test_utils
+			$<$<BOOL:${LIBM_REQUIRED}>:m>
+		)
+
 	add_executable (stdin_test tests/stdin_test.c)
 	target_link_libraries (stdin_test
 		PRIVATE
@@ -1417,6 +1446,10 @@ if (BUILD_TESTING)
 	add_test (compression_size_test_opus compression_size_test opus)
 	add_test (lossy_comp_test_ogg_opus lossy_comp_test ogg_opus)
 	add_test (string_test_opus string_test opus)
+
+	### mp3-tests ###
+	add_test (mp3_test mp3_test)
+	add_test (compression_size_test_mpeg compression_size_test mpeg)
 
 	### io-tests
 	add_test (stdio_test stdio_test)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,7 +302,7 @@ add_library (sndfile
 	src/ogg_vcomment.h
 	src/ogg_vcomment.c
 	src/nms_adpcm.c
-	src/mp3.c
+	src/mpeg.c
 	src/mpeg_decode.c
 	src/mpeg_l3_encode.c
 	src/GSM610/config.h
@@ -1208,8 +1208,8 @@ if (BUILD_TESTING)
 			$<$<BOOL:${LIBM_REQUIRED}>:m>
 		)
 
-	add_executable (mp3_test tests/mp3_test.c)
-	target_link_libraries (mp3_test
+	add_executable (mpeg_test tests/mpeg_test.c)
+	target_link_libraries (mpeg_test
 		PRIVATE
 			sndfile
 			test_utils
@@ -1447,8 +1447,8 @@ if (BUILD_TESTING)
 	add_test (lossy_comp_test_ogg_opus lossy_comp_test ogg_opus)
 	add_test (string_test_opus string_test opus)
 
-	### mp3-tests ###
-	add_test (mp3_test mp3_test)
+	### mpeg-tests ###
+	add_test (mpeg_test mpeg_test)
 	add_test (compression_size_test_mpeg compression_size_test mpeg)
 
 	### io-tests
--- a/Makefile.am
+++ b/Makefile.am
@@ -69,7 +69,7 @@ src_libsndfile_la_SOURCES = src/sndfile.c src/aiff.c src/au.c src/avr.c src/caf.
 	src/macos.c src/mat4.c src/mat5.c src/nist.c src/paf.c src/pvf.c src/raw.c src/rx2.c src/sd2.c \
 	src/sds.c src/svx.c src/txw.c src/voc.c src/wve.c src/w64.c src/wavlike.c src/wav.c src/xi.c src/mpc2k.c src/rf64.c \
 	src/ogg.c src/ogg.h src/ogg_vorbis.c src/ogg_speex.c src/ogg_pcm.c src/ogg_opus.c src/ogg_vcomment.c src/ogg_vcomment.h \
-	src/common.h src/sfconfig.h src/sfendian.h src/wavlike.h src/sf_unistd.h src/chanmap.h src/mp3.c
+	src/common.h src/sfconfig.h src/sfendian.h src/wavlike.h src/sf_unistd.h src/chanmap.h src/mpeg.c
 nodist_src_libsndfile_la_SOURCES = $(nodist_include_HEADERS)
 src_libsndfile_la_LIBADD = src/GSM610/libgsm.la src/G72x/libg72x.la src/ALAC/libalac.la \
 	src/libcommon.la $(EXTERNAL_XIPH_LIBS) -lm $(MPEG_LIBS)
@@ -222,7 +222,7 @@ check_PROGRAMS += tests/sfversion tests/floating_point_test tests/write_read_tes
 	tests/checksum_test tests/external_libs_test tests/rdwr_test tests/format_check_test $(CPP_TEST) \
 	tests/channel_test tests/long_read_write_test tests/stdin_test tests/stdout_test \
 	tests/dither_test tests/fix_this tests/largefile_test tests/benchmark tests/ogg_opus_test \
-	tests/mp3_test
+	tests/mpeg_test
 
 BUILT_SOURCES += \
 	tests/write_read_test.c \
@@ -378,8 +378,8 @@ tests_cpp_test_LDADD = src/libsndfile.la
 tests_checksum_test_SOURCES = tests/checksum_test.c tests/utils.c tests/utils.h
 tests_checksum_test_LDADD = src/libsndfile.la
 
-tests_mp3_test_SOURCES = tests/mp3_test.c tests/utils.c tests/utils.h
-tests_mp3_test_LDADD = src/libsndfile.la
+tests_mpeg_test_SOURCES = tests/mpeg_test.c tests/utils.c tests/utils.h
+tests_mpeg_test_LDADD = src/libsndfile.la
 
 # Lite remove start
 tests_dwvw_test_SOURCES = tests/dwvw_test.c tests/utils.c tests/utils.h
--- a/Makefile.am
+++ b/Makefile.am
@@ -62,26 +62,26 @@ endif
 lib_LTLIBRARIES = src/libsndfile.la
 include_HEADERS = include/sndfile.hh
 nodist_include_HEADERS = include/sndfile.h
-src_libsndfile_la_CFLAGS = $(EXTERNAL_XIPH_CFLAGS)
+src_libsndfile_la_CFLAGS = $(EXTERNAL_XIPH_CFLAGS) $(MPEG_CFLAGS)
 # MinGW requires -no-undefined if a DLL is to be built.
 src_libsndfile_la_LDFLAGS = -no-undefined -version-info $(SHARED_VERSION_INFO) $(SHLIB_VERSION_ARG)
 src_libsndfile_la_SOURCES = src/sndfile.c src/aiff.c src/au.c src/avr.c src/caf.c src/dwd.c src/flac.c src/g72x.c src/htk.c src/ircam.c \
 	src/macos.c src/mat4.c src/mat5.c src/nist.c src/paf.c src/pvf.c src/raw.c src/rx2.c src/sd2.c \
 	src/sds.c src/svx.c src/txw.c src/voc.c src/wve.c src/w64.c src/wavlike.c src/wav.c src/xi.c src/mpc2k.c src/rf64.c \
-	src/ogg_vorbis.c src/ogg_speex.c src/ogg_pcm.c src/ogg_opus.c src/ogg_vcomment.c \
-	src/common.h src/sfconfig.h src/sfendian.h src/wavlike.h src/sf_unistd.h src/ogg.h src/chanmap.h src/ogg_vcomment.h
+	src/ogg.c src/ogg.h src/ogg_vorbis.c src/ogg_speex.c src/ogg_pcm.c src/ogg_opus.c src/ogg_vcomment.c src/ogg_vcomment.h \
+	src/common.h src/sfconfig.h src/sfendian.h src/wavlike.h src/sf_unistd.h src/chanmap.h src/mp3.c
 nodist_src_libsndfile_la_SOURCES = $(nodist_include_HEADERS)
 src_libsndfile_la_LIBADD = src/GSM610/libgsm.la src/G72x/libg72x.la src/ALAC/libalac.la \
-	src/libcommon.la $(EXTERNAL_XIPH_LIBS) -lm
+	src/libcommon.la $(EXTERNAL_XIPH_LIBS) -lm $(MPEG_LIBS)
 EXTRA_src_libsndfile_la_DEPENDENCIES = $(SYMBOL_FILES)
 
 noinst_LTLIBRARIES = src/libcommon.la
-src_libcommon_la_CFLAGS = $(EXTERNAL_XIPH_CFLAGS)
+src_libcommon_la_CFLAGS = $(EXTERNAL_XIPH_CFLAGS) $(MPEG_CFLAGS)
 src_libcommon_la_SOURCES = src/common.c src/file_io.c src/command.c src/pcm.c src/ulaw.c src/alaw.c \
 	src/float32.c src/double64.c src/ima_adpcm.c src/ms_adpcm.c src/gsm610.c src/dwvw.c src/vox_adpcm.c \
 	src/interleave.c src/strings.c src/dither.c src/cart.c src/broadcast.c src/audio_detect.c \
-	src/ima_oki_adpcm.c src/ima_oki_adpcm.h src/alac.c src/chunk.c src/ogg.c src/chanmap.c \
-	src/windows.c src/id3.c src/nms_adpcm.c $(WIN_VERSION_FILE)
+	src/ima_oki_adpcm.c src/ima_oki_adpcm.h src/alac.c src/chunk.c src/chanmap.c \
+	src/windows.c src/id3.c src/id3.h src/nms_adpcm.c src/mpeg_decode.c src/mpeg_l3_encode.c src/mpeg.h $(WIN_VERSION_FILE)
 
 check_PROGRAMS = src/test_main
 src_test_main_SOURCES = src/test_main.c src/test_main.h src/test_conversions.c src/test_float.c src/test_endswap.c \
@@ -221,7 +221,8 @@ check_PROGRAMS += tests/sfversion tests/floating_point_test tests/write_read_tes
 	tests/locale_test tests/win32_ordinal_test tests/ogg_test tests/compression_size_test \
 	tests/checksum_test tests/external_libs_test tests/rdwr_test tests/format_check_test $(CPP_TEST) \
 	tests/channel_test tests/long_read_write_test tests/stdin_test tests/stdout_test \
-	tests/dither_test tests/fix_this tests/largefile_test tests/benchmark tests/ogg_opus_test
+	tests/dither_test tests/fix_this tests/largefile_test tests/benchmark tests/ogg_opus_test \
+	tests/mp3_test
 
 BUILT_SOURCES += \
 	tests/write_read_test.c \
@@ -376,6 +377,9 @@ tests_cpp_test_LDADD = src/libsndfile.la
 
 tests_checksum_test_SOURCES = tests/checksum_test.c tests/utils.c tests/utils.h
 tests_checksum_test_LDADD = src/libsndfile.la
+
+tests_mp3_test_SOURCES = tests/mp3_test.c tests/utils.c tests/utils.h
+tests_mp3_test_LDADD = src/libsndfile.la
 
 # Lite remove start
 tests_dwvw_test_SOURCES = tests/dwvw_test.c tests/utils.c tests/utils.h
--- a/cmake/FindLame.cmake
+++ b/cmake/FindLame.cmake
@@ -1,0 +1,66 @@
+# - Find lame
+# Find the native lame includes and libraries
+#
+#  LAME_INCLUDE_DIRS - where to find lame.h, etc.
+#  LAME_LIBRARIES    - List of libraries when using lame.
+#  LAME_FOUND        - True if Lame found.
+
+if (LAME_INCLUDE_DIR)
+    # Already in cache, be silent
+    set(LAME_FIND_QUIETLY TRUE)
+endif ()
+
+find_path (LAME_INCLUDE_DIR lame/lame.h
+	HINTS
+		${LAME_ROOT}
+	)
+
+# MSVC built lame may be named mp3lame_static.
+# The provided project files name the library with the lib prefix.
+
+find_library (LAME_LIBRARY
+	NAMES
+		mp3lame
+		mp3lame_static
+		libmp3lame
+		libmp3lame_static
+		libmp3lame-static
+	HINTS
+		${LAME_ROOT}
+	)
+
+find_library (LAME_HIP_LIBRARY
+	NAMES
+		mpghip-static
+		libmpghip-static
+	HINTS
+		${LAME_ROOT}
+	)
+
+# Handle the QUIETLY and REQUIRED arguments and set LAME_FOUND
+# to TRUE if all listed variables are TRUE.
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args (Lame
+	REQUIRED_VARS
+		LAME_LIBRARY
+		LAME_INCLUDE_DIR
+	)
+
+if (LAME_FOUND)
+	set (LAME_LIBRARIES ${LAME_LIBRARY} ${LAME_HIP_LIBRARY})
+	set (LAME_INCLUDE_DIRS ${LAME_INCLUDE_DIR})
+
+	if (NOT TARGET Lame::Lame)
+		add_library (Lame::Lame UNKNOWN IMPORTED)
+		set_target_properties (Lame::Lame PROPERTIES
+			INTERFACE_INCLUDE_DIRECTORIES "${LAME_INCLUDE_DIRS}"
+			IMPORTED_LOCATION "${LAME_LIBRARY}"
+		)
+		if (LAME_HIP_LIBRARY)
+			set_property (TARGET Lame::Lame APPEND PROPERTY
+				INTERFACE_LINK_LIBRARIES "${LAME_HIP_LIBRARY}")
+		endif ()
+	endif ()
+endif ()
+
+mark_as_advanced(LAME_INCLUDE_DIR LAME_LIBRARY LAME_HIP_LIBRARY)
--- a/cmake/FindMpg123.cmake
+++ b/cmake/FindMpg123.cmake
@@ -1,0 +1,64 @@
+# - Find mpg123
+# Find the native mpg123 includes and libraries
+#
+#  MPG123_INCLUDE_DIRS - where to find mpg123.h, etc.
+#  MPG123_LIBRARIES    - List of libraries when using mpg123.
+#  MPG123_FOUND        - True if Mpg123 found.
+
+if (MPG123_INCLUDE_DIR)
+    # Already in cache, be silent
+    set(MPG123_FIND_QUIETLY TRUE)
+endif ()
+
+find_package (PkgConfig QUIET)
+pkg_check_modules(PC_MPG123 QUIET libmpg123>=1.25.10)
+
+set (MPG123_VERSION ${PC_MPG123_VERSION})
+
+find_path (MPG123_INCLUDE_DIR mpg123.h
+	HINTS
+		${PC_MPG123_INCLUDEDIR}
+		${PC_MPG123_INCLUDE_DIRS}
+		${MPG123_ROOT}
+	)
+
+# MSVC built mpg123 may be named mpg123_static.
+# The provided project files name the library with the lib prefix.
+
+find_library (MPG123_LIBRARY
+	NAMES
+		mpg123
+		mpg123_static
+		libmpg123
+		libmpg123_static
+	HINTS
+		${PC_MPG123_LIBDIR}
+		${PC_MPG123_LIBRARY_DIRS}
+		${MPG123_ROOT}
+	)
+
+# Handle the QUIETLY and REQUIRED arguments and set MPG123_FOUND
+# to TRUE if all listed variables are TRUE.
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args (Mpg123
+	REQUIRED_VARS
+		MPG123_LIBRARY
+		MPG123_INCLUDE_DIR
+	VERSION_VAR
+		MPG123_VERSION
+	)
+
+if (MPG123_FOUND)
+	set (MPG123_LIBRARIES ${MPG123_LIBRARY})
+	set (MPG123_INCLUDE_DIRS ${MPG123_INCLUDE_DIR})
+
+	if (NOT TARGET MPG123::libmpg123)
+		add_library (MPG123::libmpg123 UNKNOWN IMPORTED)
+		set_target_properties (MPG123::libmpg123 PROPERTIES
+			INTERFACE_INCLUDE_DIRECTORIES "${MPG123_INCLUDE_DIRS}"
+			IMPORTED_LOCATION "${MPG123_LIBRARIES}"
+		)
+	endif ()
+endif ()
+
+mark_as_advanced(MPG123_INCLUDE_DIR MPG123_LIBRARY)
--- a/cmake/SndFileChecks.cmake
+++ b/cmake/SndFileChecks.cmake
@@ -56,6 +56,14 @@ else ()
 	set (HAVE_EXTERNAL_XIPH_LIBS 0)
 endif ()
 
+find_package (Lame)
+find_package (Mpg123 1.25.10)
+if (LAME_FOUND AND MPG123_FOUND)
+	set (HAVE_MPEG_LIBS 1)
+else ()
+	set (HAVE_MPEG_LIBS 0)
+endif()
+
 find_package (Speex)
 find_package (SQLite3)
 
--- a/configure.ac
+++ b/configure.ac
@@ -147,6 +147,9 @@ AC_ARG_ENABLE([alsa],
 AC_ARG_ENABLE([external-libs],
 	[AS_HELP_STRING([--disable-external-libs], [disable use of FLAC, Ogg and Vorbis [[default=no]]])])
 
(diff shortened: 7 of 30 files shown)


njh:
Have you seen #493 by @jpffitch ?

arthurt:
Yes, I have seen #493, it's what inspired to put this up, having sat on it for a while. I hope that this and #493 can be merged into one another.

There are a few differences between this and the approach of #493. Firstly, as encoding and decoding use completely orthogonal libraries, encoding and decoding are split into separate files.

The subformat types introduced differ. #493 introduces the subtypes SF_FORMAT_MP3_CBR and SF_FORMAT_MP3_VBR to represent . #499 (this) introduces SF_FORMAT_MPEG_I through SF_FORMAT_MPEG_III.

Otherwise, it is a measure of completeness in different areas comparing the two, and much of the code from two could be merged into one another (id3 support, xing tag support, normalization support.)

jpffitch:
My attempt has got bogged down in reading.    I failed to get mpg123 to work and I was warned off the HIP part of lame.  Currently I have been experimenting with a different library -- I seem to loose half the input but reading ints otherwise works (for some value of work...)  If you have got further I am happy to withdraw.  I will look at your PR

arthurt:
I was hoping you would take this over from me :) I too got bogged down in this. MP3 as a file-format is under-specified.

I looked at HIP from liblame too. No seek support, and questions about what it would do with non-vanilla files.

What other library are you looking at? I looked at libmad too, but again, no seek support. mpg123 sometimes seems a bit heavy.

jpffitch:
I was reasonably happy with the writing of mp3 files using the lame library.  With mpg123 it took me time to realise that it is frame oriented and I did not discover how to abstract beyond it.

The library I have been looking at recently is mpadec from a Russian I found on sourceforge which I then used in csound to allow reading of mp3 files (and encoding via pipes to lame).  I revised the sources and did a few minor fixes.  It is only used simply in csound with an opcode mp3in that reads stereo files from start to end in chunks.  There is a drawback that it is written so at open time one selects the format of the output.  I was planning to read 32bit and castthe result with range changes.  At present I am just trying to read 32 bit integers mono or stereo.  Looks feasible but it is a little known library.

What I got nowhere with was the GNU configuration tools.  Never used them for anything I have written so I just hacked cmake scripts etc to build and test.

Do you have views on merging our two schemes?  Happy to help if useful.


⁣Sent from TypeApp ​

erikd:
First task is the rebase this against current master to remove the merge commits.

Does this include tests?

jpffitch:
This version runs the mpeg_test code (simple edit from ogg_test)

    mpeg_short_test                : mpeg_short.mp3 .......... ok
    mpeg_seek_test                 : mpeg_short.mp3 .......... ok
    mpeg_int_test                  : mpeg_int.mp3 ............ ok
    mpeg_seek_test                 : mpeg_int.mp3 ............ ok
    mpeg_float_test                : mpeg_float.mp3 .......... ok
    mpeg_seek_test                 : mpeg_float.mp3 .......... ok
    mpeg_double_test               : mpeg_double.mp3 ......... ok
    mpeg_seek_test                 : mpeg_double.mp3 ......... ok
    mpeg_stereo_seek_test          : mpeg_seek.mpeg .......... ok

Will do some more tests but so far looks good.

arthurt:
Thank you @jpffitch! I have included your test program, and thanks again for finding the float decoding bug.

arthurt:
In this pull, the major format specifier is named `SF_FORMAT_MPEG` and there are three minors: `SF_FORMAT_MPEG_I` through `SF_FORMAT_MPEG_III`. Support for layers other than just three was added, because mpg123 supports all three layers, we would have to do work to exclude them.

(Only layer III is supported for writing. Nothing precludes using _twolame_ for layer II writing support at some future point.)

The more I think about it, I wonder if renaming the major format from `SF_FORMAT_MPEG` back to `SF_FORMAT_MP3` that it was before might be a good idea.

So-called mp3 files aren't actually a thing, there is no specification, just conventions on headers and trailers added to a MPEG audio layer III stream shoved in a file. `SF_FORMAT_MP3` has utility in that it makes common assumptions explicit, and models common conception about 'mp3' files.

I still think the subformat types should model MPEG audio layers.

Major | Minor |  Description
---|---|---
SF_FORMAT_MP3 | SF_FORMAT_MPEG_III | A "mp3" file. Includes Lame/Xing/Info header and id3v1 trailer, id3v2 header.
SF_FORMAT_RAW | SF_FORMAT_MPEG_* | A mpeg stream stored in a file. No Lame/Xing/Info header. Constant bit rate by default, maybe forced?
SF_FORMAT_WAV | SF_FORMAT_MPEG_III | A mpeg stream in a wav file (WAVE_FORMAT_MPEG and WAVE_FORMAT_MPEGLAYER3) It was a thing.

One case to look out for is `SF_FORMAT_MP3 | SF_FORMAT_MPEG_III`  vs `SF_FORMAT_RAW | SF_FORMAT_MPEG_III`. If one saves a `RAW|MPEG_III` file, then opens it again, it should be detected as `RAW|MPEG_III` not `MP3|MPEG_III`, as the tests would consider that an error. I think that it should be easy to differentiate, based on the lack of and Lame/Xing/Info header and/or ID3v1 trailer. This does require forcing `SF_FORMAT_MP3` to cause a Lame info header to be written even in CBR files, but I would suggest that if a user wants a CBR MP3 with no Lame/Xing/Info header, they should use `RAW|MPEG_III`.

darrenw2112:
Is this going to be merged?  MP3 support would be great.

arthurt:
Needs more work at this time sadly, and I need more time (or money, can contract, call me.) 'MP3 files' are an ad-hoc dirty standard, based on MPEG-1 and MPEG-2 sure, but there's a whole lot of fuzzy crap around the edges. Add to that the fact that the leading libraries for them, lame (lackluster API) and mpg123 (not threadsafe), and well, you have where we are now.

martinwguy:
> lame (lackluster API) and mpg123 (not threadsafe)

Why, is libsndfile thread safe? And libmp3lame uses libmpg123 for decoding anyway. However.

AngledLuffa:
What do you mean by "needs more work"?

erikd:
@martinwguy 

> Why, is libsndfile thread safe?

Yes, it is mostly thread safe. The only exception is in error handling and reporting.

martinwguy:
> What do you mean by "needs more work"?

According to https://github.com/libsndfile/libsndfile/issues/258#issuecomment-703282440 by the MP3 branch's current maintainer

AngledLuffa:
What kind of work, though?

martinwguy:
> What kind of work, though?

I'm trying to find that out.

martinwguy:
I've tried compiling against the mpeg-support branch and, for all MP3 formats I have (MPEG ADTS, layer II , v1 / MPEG ADTS, layer III, v1 and 2 / Audio file with ID3 version 2.2.0, contains:MPEG ADTS, layer III, v1) I get "File contains data in an unimplemented format" for the first three and "Format not recognised" for the last, having used
`
git clone http://github.com/arthurt/libsndfile libsndfile-arthurt
cd !$
git checkout mpeg-support
./autoen.sh
./configure --enable-mpeg
make
sudo make install
`
arthur, is there something obvious that I'm missing here?

arthurt:
> Why, is libsndfile thread safe? And libmp3lame uses libmpg123 for decoding anyway.

mpg123 has non-thread safe global init and cleanup functions. It's a problem that can be ignored for now, it's some table precalculation stuff, so threads could race at writing the same data, but it feels dirty, and who knows, maybe some future version of mpg123 will use the global functions for more.

Yes, lame contains a mp3 decoder in it's API which is mpg123, but it is an old version which has been copied into the lame codebase, and while it doesn't contain the non-threadsafe parts, also is very much lacking. (Been there, tried it.)

martinwguy:
OK, I've run some tests on this and got help privately from arthurt, so I'm forwarding the salient parts of that in case they're userful to other testers (I know, I should have kept it on-list)

Me:

I'm trying to test the new code, but with all MP3 files, sf_open
fails saying "File contains data in an unimplemented format." or
"Format not recognised.", using:

    git clone http://github.com/arthurt/libsndfile libsndfile-arthurt
    cd libsndfile-arthurt
    git checkout mpeg-support
    ./autogen.sh
    ./configure --enable-mpeg
    make
    sudo make install

Is there something obvious I'm missing here?

Arthur:

To compile with MPEG support not only do you have to pass --enable-mpeg, you also have to pass --enable-experimental.

Me:

OK! It was saying

    External MPEG Lame/MPG123 : ........... yes

anyway, so maybe that could be protected by some kind of &&
EXPERIMENTAL, so as not to fool any other fools.

Single-threaded MP3 decoding seems to work flawlessly for me. I'm
hacking my app to make the multiple FFT threads open the sound file
once each and beat libsndfile to multi-threaded death. I'll let you
know the results

Results:

The app has five threads doing essentially random seek-reads on five separate SNDFILE*'s open to the same file (one for the audio player and four for FFTs). It works fine on WAV files but, after a few seconds, on OGG files the seek fails and then it dumps core when ogg_sync_buffers() calls realloc() with an invalid pointer and with FLAC files, either the seeks start failing, or it segmentation faults or (apparently) the calls to sf_read() stop returning any data.

With multiple threads, the mpeg code wins the prize though:

    Note: Illegal Audio-MPEG-Header 0xe6000c6c at offset 8154.
    Note: Trying to resync...
    Note: Skipped 1024 bytes in input.
    [src/libmpg123/parse.c:1273] error: Giving up resync after 1024 bytes - your stream is not nice... (maybe increasing resync limit could help).
    *** stack smashing detected ***: <unknown> terminated
    Segmentation fault (core dumped)

Putting a lock()-unlock() pair around each call to sf_*() functions, to ensure that only one call is in flight at a time, cures all of these problems and everything works as it should.

I can only conclude that libsndfile's WAV reader seems thread safe, but that the ogg- and flac-reading libraries. or the way libsndfile calls them, are not, even though libflac itself does claim to be so.

One oddity: though the MP3 reader works on what "file" calls "MPEG ADTS, layer II, and layer III, v1 and v2", it says "Format not recognized" on what "file" calls "Audio file with ID3 version 2.2.0, contains:MPEG ADTS, layer III, v1". snd-file-play fails similarly but the mpg123 command plays it OK.

The MP3 reading in arthurt's mpeg-support branch seems to be stable. I haven't tested its encoding side.

arthurt:
> Putting a lock()-unlock() pair around each call to sf_*() functions, to ensure that only one call is in flight at a time, cures all of these problems and everything works as it should.

Are you sharing and simultaneously using a SNDFILE* context between threads? That is very much not supported. libsndfile is a thread safe library, as in multiple threads can use the library within one process. SNDFILE contexts are not threadsafe, a context may only be used by  one thread at a time.

> One oddity: though the MP3 reader works on what "file" calls "MPEG ADTS, layer II, and layer III, v1 and v2", it says "Format not recognized" on what "file" calls "Audio file with ID3 version 2.2.0, contains:MPEG ADTS, layer III, v1". snd-file-play fails similarly but the mpg123 command plays it OK.

That is probably an existing bug with very large ID3v2 headers in libsndfile (like such headers containing album art.) It's outstanding but not in the scope of this pull request. Check the output of `sndfile-info`:
```
Error : Not able to open input file blah.mp3.
File : blah.mp3
Length : 5137651
Found 'ID3' marker.
ID3 length : 97587
--------------------
Request for header allocation of 195194 denied.
```

martinwguy:
> Are you sharing and simultaneously using a SNDFILE* context between threads?

I understand that, but believe not. The main loop sf_opens it once,
and that it used by the audio callback, then each of the FFT threads
sf_opens the file itself and uses its own SNDFILE pointer for each of
its calls to libsndfile.

I assume that if it were doing what you say, then WAV files would not
work either, but they do.
That may be because internally libsndfile itself has been written
thread safely but has some glitches when calling external libraries. A
simple way to achieve that could be to put locks around its calls to
external libraries, but this isn't about making libsndfile thread
safe, it's about including MP3 support, whose only defect seems to be
shared by all the other external decoders.

>> One oddity: it says "Format not recognized" on
>> what "file" calls "Audio file with ID3 version 2.2.0, contains:MPEG ADTS,
>> layer III, v1".
>
> Check the output of `sndfile-info`:
```
Error : Not able to open input file test/Razzle.mp3.
File : test/Razzle.mp3
Length : 4581125

Format not recognised.
```

martinwguy:
> Are you sharing and simultaneously using a SNDFILE* context between threads?

I don't think so but will run further checks.

> > One oddity: it says "Format not recognized" on what "file" calls
> > "Audio file with ID3 version 2.2.0, contains:MPEG ADTS, layer III, v1". 
> That is probably an existing bug with very large ID3v2 headers in libsndfile
> Check the output of `sndfile-info`:

    Error : Not able to open input file test/Razzle.mp3.
    File : test/Razzle.mp3
    Length : 4581125
    
    Format not recognised.

martinwguy:
> > Are you sharing and simultaneously using a SNDFILE* context between threads?
> 
> I don't think so but will run further checks.

Yes, I'd missed one case; the audio player was using the last-opened audio file instead of its own handle. With that fixed, it works fine for WAV, OGG, FLAC and MP3 too, so libmpg123 (1.25.10) and mpeg-support appear to be thread safe. And libsndfile too of course - maybe that could be mentioned in the Features page.

Re: the oddity, calling libmpg123 does decode it correctly, but libsndfile-mpeg-support doesn't recognise it. If you're interested in looking at that, the file is under martinwguy.net/test

arthurt:
> Re: the oddity, calling libmpg123 does decode it correctly, but libsndfile-mpeg-support doesn't recognise it. If you're interested in looking at that, the file is under martinwguy.net/test

Thanks! Turns out it was because the file has a ID3v2.**2** header, whereas libsndfile would only recognise ID3v2.**3**. Fixed now.

evpobr:
@arthurt, is it not WIP now?

martinwguy:
On 12/10/2020, evpobr <notifications@github.com> wrote:
> @arthurt, is it not WIP now?

It appears to work flawlessly even under stressful conditions, and
decodes all the strange MP3s I have on hand. We are still ironing out
some minor problems, such as portabiilty to different (older) versions
of libmpg123, so I'd say it's in alpha testing now.

If you'd like to try it and see if you encounter any problems, you can
(on a Unix-like system):

    git clone https://github.com/arthurt/libsndfile
    cd libsndfile
    git checkout mpeg-support
    sh autogen.sh
    ./configure --enable-experimental --enable-mpeg
    make
    sudo make install  # Goes under /usr/local, overriding the OS's
installed one

The only outstanding question I know of is that libmpg123 says that
mpg123_init() and _exit() should be called "once in a non-parallel
context." In practice, they *are* thread-safe because _init only
initializes static tables, always to the same values, and _exit does
nothing and is declared to be a redundant stub, only kept for old
programs that still call it. See
https://mpg123.de/api/group__mpg123__init.shtml#gad59b5dc08fb7551ef5ec085906a85604

The only worry is that mpg123_init() might become non-thread-safe in future.

This could be resolved by ensuring that mpg123_init() is only called
once, but this means putting a mutex around libmpg123's call to it,
which either means making the addition of -lpthread obligatory for
everyone who uses libsndfile, or implementing a range of OS-specific
mutex calls, both of which seem far worse than the (nonexistent)
problem.

To double check, I've tried calling mpg123_init() before every read,
with 5 concurrent threads and saw that it *is* occasionally entered
more than once in parallel, and everything does continues to work
fine.

    M

arthurt:
> @arthurt, is it not WIP now?

Yes, I'd like this to be considered ready for review for inclusion. Before merging it needs to be made non-experimental so the CI builds actually test it.

Outstanding issues that I'd like to fix are not dirrectly related to the feature. Mainly, large ID3v2 headers (such as those with album art) currently don't work. (Issue with header_seek())

evpobr:
Ok, start testing.

AngledLuffa:
Any suggestions on how to test it on Windows?  I'll gladly take part if
it's easy to do.

evpobr:
It works, but needs some modifications. I'm buzy now, wait a little for updates.

evpobr:
Patch is arthurt/libsndfile#1

evpobr:
@arthurt/libsndfile#2 fails:


    ../src/mpeg_decode.c:549:67: error: use of undeclared identifier 'MPG123_NO_FRANKENSTEIN'; did you mean 'MPG123_FRANKENSTEIN'?
            mpg123_param (pmp3d->pmh, MPG123_ADD_FLAGS, MPG123_FORCE_FLOAT | MPG123_NO_FRANKENSTEIN, 1.0) ;

martinwguy:
On 24/10/2020, evpobr <notifications@github.com> wrote:
> @arthurt/libsndfile#2 fails:
>
>     ../src/mpeg_decode.c:549:67: error: use of undeclared identifier
> 'MPG123_NO_FRANKENSTEIN'; did you mean 'MPG123_FRANKENSTEIN'?
>             mpg123_param (pmp3d->pmh, MPG123_ADD_FLAGS, MPG123_FORCE_FLOAT |
> MPG123_NO_FRANKENSTEIN, 1.0) ;

This has just been addressed by
https://github.com/libsndfile/libsndfile/commit/b32d5f47
by upping the libmpg123 version from 1.25 to 1.26
though this stops it working with current Debian stable and maybe others.

I would have preferred:

--- a/src/mpeg_decode.c
+++ b/src/mpeg_decode.c
@@ -546,7 +546,11 @@ mpeg_decoder_init (SF_PRIVATE *psf)
                mpeg_dec_io_read, mpeg_dec_io_lseek, NULL) ;

        mpg123_param (pmp3d->pmh, MPG123_REMOVE_FLAGS,
MPG123_AUTO_RESAMPLE, 1.0) ;
-       mpg123_param (pmp3d->pmh, MPG123_ADD_FLAGS, MPG123_FORCE_FLOAT
| MPG123_NO_FRANKENSTEIN, 1.0) ;
+       mpg123_param (pmp3d->pmh, MPG123_ADD_FLAGS, MPG123_FORCE_FLOAT
+#if MPG123_API_VERSION >= 45
+               | MPG123_NO_FRANKENSTEIN /* Introduced in libmpg123 v1.26 */
+#endif
+               , 1.0) ;
        //mpg123_param (pmp3d->pmh, MPG123_VERBOSE, 12, 1.0) ;

        psf->dataoffset = 0 ;

   M

evpobr:
@arthurt , please rebase on master to fix failing GitHub Actions macOS builds.

arthurt:
Mac OS is now just failing because enabling experimental code brings in Speex, and Clang 12 is barfing on libspeex's `<speex.h>`!

evpobr:
> Mac OS is now just failing because enabling experimental code brings in Speex, and Clang 12 is barfing on libspeex's `<speex.h>`!

Maybe it's time to move MP3 out of experimental state?

evpobr:
Upgrading Vcpkg mpg123 port to from 1.25 to 1.26 (microsoft/vcpkg#14291). This should fix WIndows build erros.

I found that `mpg123` has CMake project and can export config-file package and version info. So `find_package(MPG123 1.26)` will work at least with Vcpkg on Windows Platform.

@arthurt , note official CMake target for `mpg123` library is `MPG123::libmpg123`.

martinwguy:
> Maybe it's time to move MP3 out of experimental state?

+1

It seems to be stable, at the few-last-quirks stage.

   M

martinwguy:
Moving it out of experimental in your branch would solve the CI issue
at least! :)

   M

evpobr:
@arthurt , i think the version 1.26 of libmpg123 is too new. `MPG123_NO_FRANKENSTEIN` must be enabled only if supported.

If you need that `MPG123_NO_FRANKENSTEIN` for something (tests?) and you cannot determine `libmpg123` version you can check `MPG123_NO_FRANKENSTEIN` presence in header, for CMake it is [CheckSymbolExists](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html) macro.

jonashaag:
I think there is a problem with the implementation: `mpg123_read` may return something other than `MPG123_OK` when decoding was successful. Namely, it may return `MPG123_DONE`: "Message: Track ended. Stop decoding."

I have an example file that demonstrates the issue.

Another issue is that in case an actual error happened inside `mpg123_read`, it will return `0`, and then `sf_read_double` will return `0` without setting any error. Shouldn't there be an error set in that case?

jonashaag:
I think there is another problem with ID3 headers which aren't skipped correctly. It only skips the 10 header bytes but not the header contents. Probably has something to do with this comment?

```c
/* TODO HACK: 'recover' the ID3v2 header. */
//psf->dataoffset = psf->fileoffset ;
```

jonashaag:
> It only skips the 10 header bytes but not the header contents

Aha, this seems to be related to `ID3 length : 212438` and `Request for header allocation of 424896 denied.`. If I change the limit everything works OK.

martinwguy:
On 04/12/2020, Jonas Haag <notifications@github.com> wrote:
> I think there is a problem with the implementation: `mpg123_read` may return
> something other than `MPG123_OK` when decoding was successful. Namely, it
> may return `MPG123_DONE`: "Message: Track ended. Stop decoding."
>
> I have an example file that demonstrates the issue.
>
> Another issue is that in case an actual error happened inside `mpg123_read`,
> it will return `0`, and then `sf_read_double` will return `0` without
> setting any error. Shouldn't there be an error set in that case?

Hi Jonas!
  Can you forward the examples that demonstrate these and the ID3
header skipping issue you mention, so I can test this and look for a
fix?
  My other question is, for each of these problems, does the "mpg123"
command manage to decode and play them correctly, as if it does, then
we have a better hope of making our use of libmpg123 follow it.
  Thanks for finding and reporting these cases. MP3 is an awkward
format as there are so many variants and odd files, and it will be a
relief when we have a solid implementation for libsndfile.

Blessings

   M

jonashaag:
Update: Found another file with an ID3 tag of size 865820 (!). These were generated by ffmpeg. Not sure why it has such large tag...

jonashaag:
> Can you forward the examples that demonstrate these and the ID3 header skipping issue you mention, so I can test this and look for a fix?

Sure, will upload here. Just to be sure, you're citing the other issue (error reporting), are you asking for samples here too?

> My other question is, for each of these problems, does the "mpg123" command manage to decode and play them correctly, as if it does, then we have a better hope of making our use of libmpg123 follow it.

Yes, these work with mpg123 and mpg123-python

jonashaag:
File with the count reporting issue. It's from Mozilla Common Voice (CC licensed). It's a mp3 file but GitHub does not allow the `.mp3` extension, so uploading with a `.zip` extension.

[common_voice_it_20042102.zip](https://github.com/libsndfile/libsndfile/files/5648992/common_voice_it_20042102.zip)

martinwguy:
On 06/12/2020, Jonas Haag <notifications@github.com> wrote:
>> Can you forward the examples that demonstrate these and the ID3 header
>> skipping issue you mention, so I can test this and look for a fix?
>
> Sure, will upload here. Just to be sure, you're citing the other issue
> (error reporting), are you asking for samples here too?

Absolutely. Being able to reproduce a bug is essential to being able
to work on it.

Thanks again

    M

arthurt:
> I think there is a problem with the implementation: `mpg123_read` may return something other than `MPG123_OK` when decoding was successful. Namely, it may return `MPG123_DONE`: "Message: Track ended. Stop decoding."
> 
> I have an example file that demonstrates the issue.
> 
> Another issue is that in case an actual error happened inside `mpg123_read`, it will return `0`, and then `sf_read_double` will return `0` without setting any error. Shouldn't there be an error set in that case?

Hi Jonas, thanks for reporting these issues!

The ID3v2 header to large size issue is an existing bug in libsndfile. I would like to address the issue, but I feel it should be in a subsequent or parallel PR. Some ID3v2 headers with album artwork can be quite large. libsndfile parses and skips the id3 header using it's binheader code. This has the effect of buffering the complete id3 header in the binheader buffer. This buffer has a limit, and some ID3v2 tag exceed it, hence the "Request for header allocation of X denied." In these cases libsndfile error's out before it ever hits a MPEG stream to detect (or Flac, or any other format you might use ID3v2 with.)

Thanks for reporting the issue of ignoring the returned value of mpg123_read(). Yup, this should be dealt with before the merge:

```
error = mpg123_read (pmp3d->pmh, ubuf.ucbuf, bufferlen * sizeof (float), &done) ;
if (error != MPG123_OK)
{	/* TODO: handle decoding error */
	break ;
	}
```

Another deficiency of the current code is streams read from a pipe currently doesn't work, but I have a fix for it.

sobukus:
Hi, mpg123 upstream here. I just stumbled upon this … so mp3 decoding is close to land in libsndfile, that is great news. Thanks for people who put work into this. If there are mpg123-specific issues, I'm happy to help out/explain things.

Also regarding libmp3lame … the LAME project is also finally moving away from their internal copy of old mpg123 code in favour of libmpg123 for decoding. Not sure if you could get away with using libmpg123 indirectly for decoding that way. I guess direct access is better for things involving sample-accurate seeking, for example. And … you're using libmpg123 in gapless mode, right? LAME + mpg123 is the team that produces proper MP3 files without the stupid padding before/after the real data. They do belong together and it's great if libsndfile can combine them under one API for simple reading and writing. The standard preset of lame works nicely enough so that you don't need to tweak encoding parameters for transparent quality.

I did not look at the code yet, but since this is for sure redundant with functionality from libsndfile itself: I hope you consider preferring mpg123's offering of differing sample encodings. Some decoders are optimized to produce f32, s16 or s32 directly. 8 bit and 24 bit are post-processed in plain C, though, so libsndfile code could be better at handling that. The next best thing would be to always start with f32 or s32 ARM/integer platforms) and convert from there. The speed benefit of direct 16 bit is mostly prominent on old platforms like MMX and 3DNow, but also somewhat on ARM.

If you want to have a consistent approach with sample conversion and ditherhing in libsndfile, f32/s32 only and then down-conversion in libsndfile seems like the way to go.

I want to point out that I too don't like the global init funcitons and want to assure you that the plan is not to extend what they do. Actually, the exit function does nothing already. I do not know how much of a nuisance they are. In a C++ project using libmpg123, I just got some static object that calls mpg123_init() in a constructor and don't think about it. There may be more elegant approaches.

As for the table calculation that mpg123_init() does, of course they could be pre-computed, but this is a bit messy because of the support for floating point and integer math. The compiled code does one or the other from the same source using a macro. When hardcoding the values, two variants have to be generated and the raw values look like black magic rather than a loop evaluating a clear formula.

But if the existence of `mpg123_init()` this is is real downer, I could be motivated to invest some time to extract the respective code and prepare pre-built headers with the table values for mpg123 1.27. The caching of decoders supported by  the current CPU would have to move to `mpg123_new()`, but that is not such a big deal.

martinwguy:
On 14/12/2020, Thomas Orgis <notifications@github.com> wrote:
> As for the table calculation that mpg123_init() does, of course they could
> be pre-computed, but this is a bit messy because of the support for floating
> point and integer math. The compiled code does one or the other from the
> same source using a macro. When hardcoding the values, two variants have to
> be generated and the raw values look like black magic rather than a loop
> evaluating a clear formula.

One could generate a .c file from a compiled program to ensure they're
correct and clearer.

    M

arthurt:
> Hi, mpg123 upstream here. I just stumbled upon this … so mp3 decoding is close to land in libsndfile, that is great news. Thanks for people who put work into this. If there are mpg123-specific issues, I'm happy to help out/explain things.

Hello!

> Also regarding libmp3lame … the LAME project is also finally moving away from their internal copy of old mpg123 code in favour of libmpg123 for decoding. Not sure if you could get away with using libmpg123 indirectly for decoding that way. I guess direct access is better for things involving sample-accurate seeking, for example.

Yes, I think continuing to use mpg123 directly rather than through lame is the best course.

> And … you're using libmpg123 in gapless mode, right? LAME + mpg123 is the team that produces proper MP3 files without the stupid padding before/after the real data. They do belong together and it's great if libsndfile can combine them under one API for simple reading and writing.

The code currently assumes gapless mpg123 support but doesn't force it. Might have to change that.
 
> I did not look at the code yet, but since this is for sure redundant with functionality from libsndfile itself: I hope you consider preferring mpg123's offering of differing sample encodings. Some decoders are optimized to produce f32, s16 or s32 directly. 8 bit and 24 bit are post-processed in plain C, though, so libsndfile code could be better at handling that. The next best thing would be to always start with f32 or s32 ARM/integer platforms) and convert from there. The speed benefit of direct 16 bit is mostly prominent on old platforms like MMX and 3DNow, but also somewhat on ARM.
> 
> If you want to have a consistent approach with sample conversion and ditherhing in libsndfile, f32/s32 only and then down-conversion in libsndfile seems like the way to go.

That is the way we have to go. Consumers of libsndfile can read x samples as s16, then read y samples as f32, then z samples as s32. (In practice most consumers will probably read the entire buffer as one sample format.) It isn't necessarily realistic to re-create the decoder and seek to were we just left off for a change of read out sample format. libsndfile also makes some guarantees about conversions from one sample format and range to another, and changing decoders might cause differences. Currently the code forces mpg123 to f32 then converts for s16, s32, f64.

> I want to point out that I too don't like the global init funcitons and want to assure you that the plan is not to extend what they do. Actually, the exit function does nothing already. I do not know how much of a nuisance they are. In a C++ project using libmpg123, I just got some static object that calls mpg123_init() in a constructor and don't think about it. There may be more elegant approaches.

Yup, I've done that too in a C++ project that needed Curl+OpenSSL, which have global init/exit requirements :-)

Things are mostly okay right now. The problem can't be solved in libsndfile, as it provides and consumes no threading functionality. It's a thread-safe-ish library, all state being tied to an opaque SNDFILE pointer, but it means that in the library we have no way to add init or exit calls.

But the way things are right now, I guess it isn't so bad.

makitsune:
I'd love to see this finished!

evpobr (approved):

sobukus:
About your id3 skipping … do I parse that correctly that you have id3_skip() only for the  case of some random non-MPEG file that has ID3v2 in front? I wonder if I should make the ID3v2 parser in libmpg123 accessible for such purpose … but then, there are proper metadata parser libraries out there.

arthurt:
Yes, the ID3 skipping is existing functionality for supporting an ID3v2 header on other formats. Currently the only practical use case I can imagine is FLAC with ID3 in front.

Note that some ID3 libraries put a buffer of zeros after the ID3 header, not accounted for in the ID3 header count field to provide space to grow. MP3 decoders are usually tolerant of this, the standard being written around sync frames in the rough.

Current state of master of is if an ID3 header is seen, the length of the ID3 header is parsed, and the header is seeked over, and the file is treated as being embedded.

Pros:
 - Uniform for all file formats

Cons:
 - Doesn't decode ID3 payload
 - Doesn't work on pipes (can't lseek a pipe, even forwards. Could work around by eating bytes to skip)
 - Will fail to identify files that have a buffer between the ID3 header and file proper.

The current state of this pull request is when an ID3 header is seen, try and open it as an MP3 file.

Pros:
 - ID3v2 support for mp3 files through libmpg123, including from pipes.
 - libmpg123 can possibly leverage metadata from ID3v2 tags for decoder setup (eg TLEN for a non-seekable pipe possibly, but I don't think this is implemented currently)
 - libmpg123 will seek through a buffer after the ID3 header for a few kb looking for a MPEG frame sync, increasing change of correctly identifying mp3 files.

Cons:
 - In the case of ID3 on a non-mp3 file, libmpg123 searches fruitlessly through the first few kb of the file for a frame seek, then we have to seek back to where we were, skip the id3 header and try again. This will not work on pipes.

I did write a trial ID3 parser using libmpg123 when available, and it did work.

Pros:
 - ID3 support uniformly for all formats
 - Supports decoding ID3 from pipes for all formats.

Cons:
 - A buffer between the end of the ID3 header and the file proper will cause the file not to be identified. Could be worked around by eating any 0-bytes for a few kb after an ID3 header before considering the ID3 header 'done.'
 - Requires opening, creating, initializing a mpg123 decoder twice.
 - Any metadata from the ID3v2 header that could be useful to libmpg123 is lost.

makitsune:
I think mp3 needs to be added to `major_formats` and `simple_formats` in `command.c`

makitsune:
I've been having issues loading in a couple of mp3 files with this code. I can provide the link to one, which when parsed with`sf_open`, `sf_strerror` says: `Error in WAV/W64/RF64 file. Malformed 'fmt ' chunk.`

I'm not sure what makes this file different but here's the mp3: https://www.myinstants.com/media/sounds/untitled_17.mp3

I've tried pre-specifying the SF_INFO format to SF_FORMAT_MP3 before opening, but that doesn't fix it. It works in every other program.

darrenw2112:
Any idea when this will make it into the master branch (a lot of great work here and mp3 support would be awesome)?

martinwguy:
On 11/01/2021, darrenw2112 <notifications@github.com> wrote:
> Any idea when this will make it into the master branch (a lot of great work
> here and mp3 support would be awesome)?

Well, Eric tends to take a long time to find the time necessary to
make a new release (7 years in my case of sndfile-tools!), and a new
release will then take several years to make it into the mainline
distributions, so I wouldn't hold your breath if I were you.

In the meantime, if you want to use the new functionality in your open
source software product, your best option is to include the source
tree of libsndfile+mpeg as a subtree of your own source code.

     M

darrenw2112:
To build this new library, you mentioned  libsndfile+mpeg.  Which branch is the one to use (mpeg-support, mpeg-decode, or mp3-codec).  A while back I saw these build instructions, are they still valid?
`git clone https://github.com/arthurt/libsndfile
cd libsndfile
git checkout mpeg-support
sh autogen.sh
./configure --enable-experimental --enable-mpeg
make
sudo make install`

martinwguy:
That's the one.

arthurt:
> I've been having issues loading in a couple of mp3 files with this code. I can provide the link to one, which when parsed withsf_open, sf_strerror says: Error in WAV/W64/RF64 file. Malformed 'fmt ' chunk.

> I'm not sure what makes this file different but here's the mp3: https://www.myinstants.com/media/sounds/untitled_17.mp3

The file linked is actually a WAV file with a MPEG audio layer 3 payload (don't let the .mp3 extension fool you ;) )

MP3 files with junk or extra data before the MPEG layer 3 audio payload are not generally considered malformed. MP3 decoders usually scan the first few kilobytes of a file for a MPEG frame sync, ignoring data before the frame sync that they don't understand as junk or unsupported headers.

MP3 files don't really exist. The only standard is the layer 3 audio codec from the MPEG-1 and MPEG-2 specifications. Everything else, including how to store it in a file, is convention or optional third-party standards.

WAV files can contain audio data of a multitudes of standards, including MPEG audio using [MPEG1WAVEFORMAT](https://docs.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-mpeg1waveformat) and [MPEGLAYER3WAVEFORMAT](https://docs.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-mpeglayer3waveformat).

However, not all software that can read WAV files support all of the possible audio formats that can be put in WAV, including support for MPEG audio in WAV (possibly even if the same software supports MP3 separately.)

Meanwhile, because MP3 decoders generally skip over junk, the WAV RIFF header is ignored as junk and the MP3 decoder will plays a MPEG-in-WAV file just fine. In software that identifies file types primarily by filename extensions, rename the file to `.mp3` and things "just work".

libsndfile tries to identify files first by searching for a header it understands, ignoring the filename. Only if that fails does it try to identify a file using the filename extension. So, when libsndfile tries to open a MPEG-audio-in-WAV file, such as the one you linked, it will successfully detect a WAV header and proceed to try and open it as a WAV file. However, MPEG1WAVEFORMAT and MPEGLAYER3WAVEFORMAT are not (yet) supported and it will report that it cannot open the file.

Adding MPEG1WAVEFORMAT and MPEGLAYER3WAVEFORMAT are on my TODO list. Once done, the file you linked to will have a format type of `SF_FORMAT_WAV | SF_FORMAT_MPEG_LAYER_III`.

> I've tried pre-specifying the SF_INFO format to SF_FORMAT_MP3 before opening, but that doesn't fix it. It works in every other program.

`sf_open` ignores the passed `SF_INFO` structure completely, except for the case of `.format` is `SF_FORMAT_RAW`.

arthurt:
>git checkout mpeg-support
sh autogen.sh
./configure --enable-experimental --enable-mpeg
make
sudo make install`

The configure flag `--enable-experimental` is no longer required :)

makitsune:
Awesome. Thanks for explaining

evpobr:
@arthurt , can you resolve conflicts?

arthurt:
> can you resolve conflicts?

Have done, haven't pushed.

Really, this should be merged BUT for the id3 handling. Hoping to resolve this soon, when the universe grants me more time.

arthurt:
Thanks for the approvals. Heads up, would like to merge.

evpobr:
> Thanks for the approvals. Heads up, would like to merge.

Good! So merge it! :smile:.

evpobr:
Many thanks to @arthurt  for this wonderful pull request!

I will update documentation.

@SoapGentoo , i guess we need to bump API version after it. Maybe 1.1.0-beta.1? And bump (patch) ABI too.

evpobr:
@arthurt , MSVC CI test builds are failing:

I've fixed CMake build (they were built without MP3 support), but something strange appears for MSVC builds:

```
2021-03-26T08:03:48.2405904Z D:\a\libsndfile\libsndfile\src\mpeg_decode.c(567,5): error C2018: unknown character '0x40' [D:\a\libsndfile\libsndfile\build\sndfile.vcxproj]
```

evpobr:
Probably `libmpg123` on Windows platform built with Vcpkg has invalid `MPG123_API_VERSION`.

evpobr:
Yep, '0x40' character is '@' and MSVC build of libmpg123 includes `mpg123.h.in` (!!!!), with:

```c
#define MPG123_API_VERSION @API_VERSION@
```

SoapGentoo:
@evpobr want me to bump it now?

evpobr:
Yes please.

SoapGentoo:
@evpobr the hypen in `1.1.0-beta.1` causes issues for the linker script:

```sh
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld:../../src/Symbols.gnu-binutils:3: ignoring invalid character `-' in script
```
Here's my suggestion: Let's use the [Xorg versioning scheme](https://www.x.org/releases/X11R7.7/doc/xorg-docs/Versions.html) (which is very common for X/GNOME/Meson and doesn't require alphabetical parsing), which means I'd bump this release to `1.0.900`, we can do patch-level bumps for multiple beta releases, and then we do the final bump to `1.1.0` once we do the release.

evpobr:
> @evpobr the hypen in 1.1.0-beta.1 causes issues for the linker script:

Oh, it is problem. Don't bump yet.

sobukus:
Note about the MSVC libmpg123 build: It is using mpg123.h.in like this:
```
#define MPG123_NO_CONFIGURE
#include "mpg123.h.in" /* Yes, .h.in; we include the configure template! */
```

But the idea is that you can use ports/cmake to create updated MSVC builds. I hope to be able to remove the MSVC build some time. The only use (for me) of the cmake build is to replace manually-curated MSVC projects. I can at least somewhat test that one on Linux, from time to time.

sobukus:
So far nobody complained about the bad API version. But I see that it has been forgotten to handle that in some sane manner. Well, another reason to drop it.

evpobr:
@sobukus , working on mpg123 vcpkg port. I hope they fixed CMake toolchain.

evpobr:
> @evpobr the hypen in `1.1.0-beta.1` causes issues for the linker script:
> 
> ```shell
> /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld:../../src/Symbols.gnu-binutils:3: ignoring invalid character `-' in script
> ```
> 
> Here's my suggestion: Let's use the [Xorg versioning scheme](https://www.x.org/releases/X11R7.7/doc/xorg-docs/Versions.html) (which is very common for X/GNOME/Meson and doesn't require alphabetical parsing), which means I'd bump this release to `1.0.900`, we can do patch-level bumps for multiple beta releases, and then we do the final bump to `1.1.0` once we do the release.

It's worse than I thought.

The only sensible way out is not to use alpha, beta, and so on, just the three basic digits. Make a bump version just before the release.

SoapGentoo:
...or just use the Xorg versioning scheme (which uses a three-digit-tuple) in the way I suggested.

evpobr:
> ...or just use the Xorg versioning scheme (which uses a three-digit-tuple) in the way I suggested.

The version with 1.0.900 is semantically incorrect. We have a minor feature release, but it looks like a patch.

evpobr:
@arthurt , I can fix the Vcpkg port, but it doesn't solve the problem with Windows in general. There are other package managers, after all libmpg123 can be built manually.

Maybe we should try to fix the problem on our side? Сan we use `#ifdef MPG123_NO_FRANKENSTEIN` instead of `#if MPG123_API_VERSION`?

evpobr:
> Maybe we should try to fix the problem on our side? Сan we use `#ifdef MPG123_NO_FRANKENSTEIN` instead of `#if MPG123_API_VERSION`?

Added in 16251b2.

All CI builds are OK: https://github.com/libsndfile/libsndfile/runs/2207364189

sobukus:
Wait, what? You're doing ifdef on an enum value. That should always be false in my book (correct me if I'm wrong). Can't we just agree on dropping support for the outdated MSVC build?

Our normal process for building the Windows binaries of mpg123 (and libmpg123) is using MinGW and autotools. This generates proper headers. That's what you can fetch from https://mpg123/download/win32/ and https://mpg123/download/win32/. The MSVC project was for people doing their own projects using libmpg123 and apparently the issue of undefined version never cropped up.

If the frankenstein flag is the only thing you're needing, another way to get out of the situtation is to just use its value:

```
        ,MPG123_NO_FRANKENSTEIN = 0x1000000 /**< Disable support for Frankenstein streams
```
MPG123_ADD_FLAGS doesn't check for invalid values, but even if it might do that in future, you're not error-checking the mpg123_param() calls anyway … It's not clean to use the value without name, and I recommend clearly marking that in a comment, but it's rather safe as the value cannot change without breaking ABI anyway.

evpobr:
> Wait, what? You're doing ifdef on an enum value.

Oh...

evpobr:
> If the frankenstein flag is the only thing you're needing, another way to get out of the situtation is to just use its value:

Thanks for that.

> Our normal process for building the Windows binaries of mpg123 (and libmpg123) is using MinGW and autotools. This generates proper headers. That's what you can fetch from https://mpg123/download/win32/ and https://mpg123/download/win32/. The MSVC project was for people doing their own projects using libmpg123 and apparently the issue of undefined version never cropped up.

Release versions of libsndfile are built in Visual Studio and use static versions of external libraries to avoid dependencies, as is common on Windows. That's why I chose Vcpkg, which can do this.

I have already done a Vcpkg port update. Now it reads `API_VERSION` from configure.ac and configures `mpg123.h.in`. As a matter of fact, I adapted the code from your CMake project. :smile: That's enough to make it work.

Don't worry, your CMake project is getting better with each release, and I'm already working on integrating it into Vcpkg.

evpobr:
sndfile.pc is outdated. @SoapGentoo , can you help?

SoapGentoo:
@evpobr what do you want me to do? Cut `1.0.900`?

evpobr:
Let's make 1.1.0. The only stopper is sndfile.pc.

SoapGentoo:
You want to cut the final `1.1.0` now?

evpobr:
@SoapGentoo , wait a little.

@arthurt, a have problems decoding MP3:

```
.\sndfile-convert.exe .\stossgebet.mp3 stossegebet.aif

Error : output file format is invalid.
The 'AIFF' container does not support 'unknown' codec data.
Run 'D:\temp\endswap_test\sndfile-convert.exe --help' for clues.
```

evpobr:
@SoapGentoo , decoding MP3 works for you?

SoapGentoo:
```sh
$ ./programs/sndfile-convert ~/Downloads/file_example_MP3_700KB.mp3 test.aif

Error : output file format is invalid.
The 'AIFF' container does not support 'unknown' codec data.
Run 'sndfile-convert --help' for clues.
```

arthurt:
I was away from the internet all weekend.

> `Error : output file format is invalid.`
> `The 'AIFF' container does not support 'unknown' codec data.`

That's not a bug with MPEG decoding. There is a bug in that the subformat name isn't being returned. What's happening is that by design sndfile-convert tries to use the same subformat. The command, as you typed it, is trying to put MPEG audio in AIFF file, which is not possible. Try the same with Opus or Ogg and it will also fail with the same error message, albeit without the 'unknown' bug.


> Let's make 1.1.0. The only stopper is sndfile.pc.

That's a bad idea. I would NOT release this as 1.1.0 until a few weeks. There are bound to be bugs with this and you'll be pushing out 1.1.1 in a bit. 1.0.900 makes it more clear that this isn't quite stable in my opinion.

> If the frankenstein flag is the only thing you're needing, another way to get out of the situtation is to just use its value:
>> MPG123_NO_FRANKENSTEIN = 0x1000000 /**< Disable support for Frankenstein streams

Using raw magic numbers is massive code stink. Like, first-year level stink. DO NOT DO THIS. Please revert the commit that does that. I would not have passed it for review. It would be preferable to not even use the flag over that.

Why are we making arch decisions based around the worlds worst platform for build systems? Aka, the Windows ecosystem.

sobukus:
Btw. I killed the MSVC build in svn trunk. You need to go through CMake for MSVC in future mpg123, and that should have correct version in the header. And sure, it's bad form to hardcode raw numbers, but if you really need the feature, it's the dirty way that works even with broken headers. If you decide you don't need to support the broken headers, all the better.

In other news … I'm also killing mpg123_init() for 1.27.0, precomputing those tables. So mpg123_init() and mpg123_exit() will soon be no-ops and you got no worry about thread safety when separating threads using mpg123 handles, just like libsndfile itself. For old versions of libmpg123 you can pretend that they're no-ops, as they don't really step on each other's toes.

evpobr:
Ok, thanks for information.

evpobr:
@SoapGentoo , @arthurt , which versioning scheme do we choose in the end - XOrg (1.0.99.1) or @erikd (1.1.0beta1)?

SoapGentoo:
I would prefer XOrg

evpobr:
> First, sorry for the intensity of my last comment.

Never mind, we are all human after all.

evpobr:
> I think X.org uses unstable releases as odd minors, which we aren't doing. So the prelease is x.<odd>.900. Humm.

I'm not sure: https://en.wikipedia.org/wiki/X_Window_System#Release_history.

@SoapGentoo , what do you think?

SoapGentoo:
That's the entire "X Window System", which is not released monolithically anymore.

https://www.x.org/releases/individual/xserver/

evpobr:
I see odd and even stable minor versions:

xorg-server-1.1.0.tar.bz2
xorg-server-1.2.0.tar.bz2
xorg-server-1.3.0.0.tar.bz2
xorg-server-1.4.tar.bz2
xorg-server-1.5.0.tar.bz2
xorg-server-1.6.0.tar.bz2

evpobr:
To be honest, I'm completely confused with this XOrg versioning scheme.

@SoapGentoo 

Our next minor prerelease version using the XOrg scheme is 1.0.99.1 or 1.0.900?

> X.org uses unstable releases as odd minors, which we aren't doing. So the prelease is x.<odd>.900. Humm.

Is this correct?

SoapGentoo:
In that case, I'd consider `1.0.99.1` to be more suitable

Nelson-numerical-software:
semantic versioning should be  really considered (https://semver.org/)

SoapGentoo:
FYI, Xorg versioning is perfectly in line with SemVer.

sobukus:
So, to be precise, SemVer would be 1.1.0-beta.1, right? I'm following this as interested outsider. The extra dot irritates me a bit, but I recognize that it eases sorting. personally, I used the scheme 1.1.0-beta1 before and think that it also shouldn't be troublesome to do this without  the dot between beta and 1 — as long as you don't do more than 9 rounds of beta/release candidate, which is sensible, IMHO.

I also used 1.2.3-somepatch in the past to indicate an addition to the base version and relied on stuff parsing that as equivalent to 1.2.3 at least… but did not expect any automatism to decide that 1.2.3-foo is newer or older than 1.2.3.

evpobr:
> So, to be precise, SemVer would be 1.1.0-beta.1, right?

Correct.

> I also used 1.2.3-somepatch in the past to indicate an addition to the base version and relied on stuff parsing that as equivalent to 1.2.3 at least… but did not expect any automatism to decide that 1.2.3-foo is newer or older than 1.2.3.

SemVer supports prereleases only, so any 1.2.3-foo is older than 1.2.3.

evpobr:
@SoapGentoo
>    /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld:../../src/Symbols.gnu-binutils:3: ignoring invalid character `-' in script

I'm pretty sure it is bug in our `create_symbols_file.py` symbol generation script.

evpobr:
> FYI, Xorg versioning is perfectly in line with SemVer.

It is not, check here https://jubianchi.github.io/semver-check

Funnily enough, such a number does not appear to be incorrect, but parsing into separate components fails. Check 1.1.0-beta.1 to see how it works.

To begin with, SemVer does not use the fourth digit of the version at all. Something similar is 'prerelease': 1.1.0-1.

New feature requires minor bump. XOrg's `1.0.99.1` doesn't reflect it,

evpobr:
I am not saying that I am not ready to accept XOrg versioning. But this is not SemVer. And we can even say that this is not entirely semantic versioning in cases of prereleases.

SoapGentoo:
Honestly, just decide on something, I'm kinda tired of this debate.

nwrkbiz:
use 1.0.99.1

evpobr:
Sorry, sometimes I can be very boring. Remind me about it :smile:

We choose the XOrg versioning scheme with the majority of votes.

evpobr:
@SoapGentoo, it's your turn now.

SoapGentoo:
so `1.0.900` for SemVer compatibility (consider how you want to handle the versioned library SONAME)?

evpobr:
> so `1.0.900` for SemVer compatibility (consider how you want to handle the versioned library SONAME)?

If I understand correctly, you are asking about the ABI version? If so, it has its own versioning scheme.

1.0.31 -> 1.0.32.

evpobr:
@arthurt , i've fixed mpg123 Vcpkg port and `MPG123_NO_FRANKENSTEIN` is now restored by #725.

darrenw2112:
Would 1.1.0beta1 be the first official release with MP3 support?

evpobr:
> Would 1.1.0beta1 be the first official release with MP3 support?

No, it is pre-release. Not all of the planned features of official 1.1.0 release have been implemented, but it is quite stable.

evpobr:
@arthurt, is there an approximate date?

nvishnoiMW:
Just to confirm: does the beta version https://github.com/libsndfile/libsndfile/releases/tag/1.1.0beta1 have MP3 read/write support?

evpobr:
> Just to confirm: does the beta version https://github.com/libsndfile/libsndfile/releases/tag/1.1.0beta1 have MP3 read/write support?

Yes.

nvishnoiMW:
> > Just to confirm: does the beta version https://github.com/libsndfile/libsndfile/releases/tag/1.1.0beta1 have MP3 read/write support?
> 
> Yes.

Thanks @evpobr!

PR #507, opened 2020-01-26
src/ima_adpcm.c: fix wavlike_ima_seek() crash

dragos-oancea:
I am not quite sure decode_block should be even called there, however this avoids the crash. 
aiff_ima_seek() may have the same problem.


erikd:
Would I be able to get a copy of the file that causes this crash?

dragos-oancea:
It crashes when writing / recording. 
https://github.com/signalwire/freeswitch/blob/master/src/mod/formats/mod_sndfile/mod_sndfile.c
The module handles well all formats, except adpcm.

erikd:
Ah ok. I spent many many CPU hours fuzzing decoding. Maybe i should spend CPU cycles on encoding as well.

dragos-oancea:
great project, btw !

PR #512, opened 2020-02-26
Set OGG page latency for Opus file.

Vahlar:
Hi, 
I would like to add a new command for the sf_command() function.
This command let the user set the OGG page latency when writing an Opus file.
This command would be very important for my work.
Thanks for your time.

C
sf_command(SNDFILE *sndfile, SFC_SET_OGG_PAGE_LATENCY_MS, void *data, int datasize);

C++
sndfile.command (SFC_SET_OGG_PAGE_LATENCY_MS, void *data, int datasize);

Max average latency set at 1600ms, min average latency set at 50ms.
default latency set at 1000ms.


mipac:
very useful!!

evpobr:
@Vahlar, just curious, why we have **two** commands in public header `sndfile.h` - `SFC_SET_OGG_PAGE_LATENCY_MS`  and `SFC_SET_OGG_PAGE_LATENCY`?

IMHO only `SFC_SET_OGG_PAGE_LATENCY_MS` is accepted by `sf_command()` function.

PR #531, opened 2020-07-04
Update Azure Pipelines CI images

evpobr:
Some test images were deprecated causing test build errors.

Visual Studio 2015 was removed and replaced with Visual Studio 2019.

macOS 10.13 was deprecated, replaced with macOS 10.15.


evpobr:
@erikd ?

PR #532, opened 2020-07-09
Update m4/add_fortify_source.m4 to fix MinGW bug

evpobr:
See source for details. Related to issue https://github.com/msys2/MINGW-packages/issues/5803.


evpobr:
This doesn't work, as workaround use:

    ./configure --enable-stack-smash-protection

evpobr:
Fixed by 16b089a.

PR #538, opened 2020-08-04
CMake fixes

evpobr:
* Set static runtime by default for MinGW

  Autotools does the same. Also fixes bug - compiler flags for MinGW were
  cleared by when ENABLE_STATIC_RUNTIME is OFF (default).

  Also ENABLE_STATIC_RUNTIME option disabled for MinGW.

* Better CMake policies handling
* Add SndFile::sndfile alias library
* Don not add programs to package config-file
* Update README.md
* Fix generator expression with src/version-metadata.rc


cmake/SndFileChecks.cmake
```diff
@@ -213,6 +213,13 @@ if (MSVC)
 	add_definitions (-D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE)
 endif (MSVC)
 
+if (MINGW)
+	set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -static-libgcc")
+	set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -static-libgcc -static-libstdc++")
+	set (CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_C_FLAGS} -static-libgcc -s")
+	set (CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS "${CMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS} -static-libgcc -static-libstdc++ -s")
+endif ()
```
SpaceIm, line 221: @evpobr Maybe I should open an issue. I would like to know why static link with libgcc and libstc++ is forced? If there is no special reason, this decision should be left to consumers.

evpobr: And it is done: libsndfile/libsndfile#630.

Issue #555, opened 2020-08-21
Possible issue with Opus pre-skip


xelsa:
Hi devs,

My application works with Ogg Opus files and I am using this amazing library to read samples. But I see that I get some bad initial samples from `sf_read_float` (I think they are zero), which causes my application to trigger an assertion later when it tries to do some work with those samples, showing that the samples are bad.

Weirdly, if I convert my Opus buffer to a WAV or FLAC buffer using `sf_write_float` and then read samples from it, the application runs perfectly normally and gives expected results.

Digging around I found the existence of `pre-skip` for Opus files. I suspect that there is some inconsistency in the behavior of the library, could it be that the library is not discarding the `pre-skip` samples and returning them?

Thanks!

arthurt:
Looking at the files, yes `toshiba-ffmpeg.ogg` would appear to have an incorrect prekip value in it's header, an thus be a malformed file.

Both `toshiba-ffmpeg.ogg` and `toshiba-opusenc.ogg` have the same total frame count when excluding the preskip, but the ffmpeg version has some extra 208 frames @ 48kHz more at the beginning, leading me to believe the preskip is wrong.

This would appear to be a bug in ffmpeg, as the erroneous preskip value of 104 is 312/3, where 3 is because of the input sample rate conversion of 48000/16000. This would suggest that ffmpeg is filling out the preskip field as frames at the input sample rate, whereas the OggOpus spec says preskip should be expressed as samples at 48kHz.

@xelsa For the other file, `split-008.ogg`, I cannot find an issue with it. What sort of error are you getting? Another issue I discovered is the way the Opus codec code reports normalizing, eg:

```
$ sndfile-info test.flac
...
Signal Max  : 31618 (-0.31 dB)

$ sndfile-info test.opus
...
Signal Max  : 1.08635 (-89.59 dB)
```

Are you relying on SFC_CALC_NORM_SIGNAL_MAX?

arthurt:
Looking at ffmpeg sources, yes it has a bug where the preskip amount is written at encode relative to the input sample rate. Worse, fmpeg always decodes Opus at 48kHz and interprets the preskip at 48kHz, so the bug is not symmetric for encode/decode within the project.

evpobr:
Interestingly, libav also have this bug?

xelsa:
Thanks for taking a look @arthurt! Yes, `split008.ogg` is where the real problem is.

[...]

PR #556, opened 2020-08-21
Opus: Fix preskip on streams with non-zero starting granule positions

arthurt:
Previously header.preskip was compared against pkt_pos, which might have an offset. Introduce gp_preskip, the granule position after the preskip amount, to make things both more explicit and removing a few instances of (gp_start + header.preskip).

~~Possibly related to issue #555.~~

### DIFF
--- a/src/ogg_opus.c
+++ b/src/ogg_opus.c
@@ -242,6 +242,7 @@ typedef struct
 			OpusMSDecoder *state ;
 			uint64_t gp_start ;
 			uint64_t gp_end ;
+			uint64_t gp_preskip ;
 			sf_count_t last_offset ;
 		} decode ;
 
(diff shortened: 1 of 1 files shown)


evpobr:
@arthurt , are you still working on this?

arthurt:
I left this as a draft as I have yet to craft an opus file with offsets to test it. The files from #555 are not offset, so this is not actually related to that issue. I'll get back to you as soon as I have tested it.

arthurt:
Update on this. I made the test, and this fix works, but so does the original should-be-broken code. I need to understand why before before progressing.

For anyone else wanting to test, you just need an ogg file with an opus stream with an initial granule position that is greater than zero.

PR #578, opened 2020-08-28
Disable extra libraries to get ossfuzz working in the ossfuzz pipelines

cmeister2:
It looks like in the final image that ossfuzz uses now, it doesn't have access to installed dynamic libraries.

To work around this for now (and get some testing done, albeit not exhaustive), disable all the extra libraries in the ossfuzz build for now. To include these later, they can be linked statically.


cmeister2:
(this should allow https://github.com/google/oss-fuzz/pull/4388 to continue)

erikd (approved):

PR #590, opened 2020-09-01
OSS-Fuzz - enable building in Github Actions CI

cmeister2:
This _should_ build the OSS-Fuzz image in CI :smile:


cmeister2:
Looks promising from https://github.com/erikd/libsndfile/pull/590/checks?check_run_id=1058772913

cmeister2:
Added in a little tweak so that it verifies the fuzzer that it builds as well.

Issue #581, opened 2020-08-30
opus test failures on BE platforms


SoapGentoo:
Hi @erikd 
we're seeing test failures on big-endian platforms in Gentoo:
```
    float_scaled_test              : opus.opus ............... 

Line 237: sf_open (SFM_READ) failed : File contains data in an unimplemented format.

File : opus.opus
Length : 2483
Ogg stream data : Opus
Stream serialno : 818667452
Opus library version: libopus 1.3.1
Opus : Unknown / unsupported embedding scheme version: 16777216.

FAIL: tests/pedantic-header-test.sh
```
with the 1.0.29 release

evpobr:
@arthurt ?

arthurt:
Looking

arthurt:
https://bugs.gentoo.org/719020

SoapGentoo:
correct, I'm reporting that bug

arthurt:
Yup, just linking it here for reference :-)

Took me a while to get a BE sandbox to play in, but yes, it is an endian issue, but more generally an integer pointer aliasing issue when parsing OggOpus header packets. Testing, but I should have a PR today to fix it.

SoapGentoo:
@arthurt thanks, tested it on our SPARC machine, tests pass now :+1:

PR #600, opened 2020-09-08
Opus: Fix integer bug in header parsing. Fixes issue #581.

arthurt:
Fix errors in parsing an OggOpus header packet where aliased pointers of different type widths are used with psf_binheader_readf(), resulting in incorrect data or endian issues. Telling psf_binheader_readf() to read an integer of fixed width, but then passing a pointer to an integer of a different width is a bug.


evpobr:
Thanks @arthurt !

PR #603, opened 2020-09-10
[Win32] Fix for MSVC

gvanem:
* A `_MSCVER` has never been a built-in for the MSVC compiler.
* Simply use the same `SF_COUNT_MAX` for all MSVC compilers.


evpobr:
Thanks, @gvanem !

PR #604, opened 2020-09-10
[Win32] fixes for MinGW

gvanem:
Most of the `S_IRxy` values could be defined in most MinGW-distros.

If MinGW's `<sys/stat.h>` gets included ahead of `sf_unistd.h`, do some `#ifndef` on these `S_IRxy` values first.
(but since `HAVE_DECL_S_IRGRP` should be 1, it would be unlikely),

### DIFF
--- a/src/sf_unistd.h
+++ b/src/sf_unistd.h
@@ -60,17 +60,39 @@
 #define		S_IXUSR	0000100	/* execute/search permission, owner */
 #endif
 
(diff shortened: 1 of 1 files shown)


evpobr:
Good, MinGW defines all of that.

Align define values using tabs, include `stat/stat.h` please.

evpobr:
Closed by #606.

PR #607, opened 2020-09-10
[Win32] support both '\\' and '/' in path-names

gvanem:
To get a working `sndfile-regtest.exe` for Windows, the assumption of 
only `'/'` on paths caused strange and non-intuitive errors like: `Not a valid SNDFILE* pointer.`

And BTW, since the Windows-Kit has `winsqlite3.h`, this could be used instead of `sqlite3.h`. Later.


gvanem:
I see now that my patch has problems with relative paths. If there is any use for this program, I'll add another commit.

evpobr:
Don't spend time on this app, it is not part of libsndfile distribution. Or you really need it?

Mark pull request as [draft](https://github.blog/2019-02-14-introducing-draft-pull-requests/) next time if it is not ready yet.

evpobr:
@gvanem , thanks!

PR #638, opened 2020-10-19
Fix memory leak in wav_read_acid_chunk

cmeister2:
Credit to OSSFuzz


evpobr:
Thanks @cmeister2 :smile:

PR #640, opened 2020-10-22
Fix up the fuzzer so that it can't under or overseek.

cmeister2:
Instead, print out the bad seek, e.g.

```
Running: /out/clusterfuzz-testcase-minimized-sndfile_fuzzer-4838501290606592
vf overseek: new_offset(539033364) > vf->length(20);  whence(1), vf->offset(20), offset(539033344)
vf overseek: new_offset(539033364) > vf->length(20);  whence(1), vf->offset(20), offset(539033344)
```

### DIFF
--- a/ossfuzz/sndfile_fuzzer.cc
+++ b/ossfuzz/sndfile_fuzzer.cc
@@ -3,6 +3,7 @@
 #include <string.h>
 #include <sys/types.h>
 #include <sndfile.h>
+#include <inttypes.h>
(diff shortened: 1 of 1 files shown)


evpobr:
I seems  like e44d327 is buggy.

cmeister2:
I didn't run the patch you sent me against the CI tests, just against the fuzzer. Might be more complicated to fix then 😢

PR #652, opened 2020-11-08
Add API version

evpobr:
@erikd, @SoapGentoo, @arthurt 

We have ABI version now, but not API version. As a result, it is difficult to determine when we added a new API. I suggest separating the API version from the ABI version.

If we add MP3 support to the next release, the API version will be 1.1.0 (API compatible changes), and the ABI will be 1.0.31 (without functional changes). More info [here](https://semver.org/).


configure.ac
```diff
@@ -85,9 +85,9 @@ dnl      age.
...
 
-CLEAN_VERSION=`echo $PACKAGE_VERSION | $SED "s/p.*//"`
-VERSION_MAJOR=`echo $PACKAGE_VERSION | $SED "s/\..*//"`
-VERSION_MINOR=`echo $CLEAN_VERSION | $SED "s/.*\.//"`
+CLEAN_VERSION="1.0.30"
+VERSION_MAJOR="1"
+VERSION_MINOR="30"
 GEN_TOOL="autoconf"
 
 SHARED_VERSION_INFO="$VERSION_MAJOR:$VERSION_MINOR:0"
```
SoapGentoo, line 93: as an aside, libtool's versioning doesn't actually work like this, it's more subtle: https://flameeyes.blog/2009/04/22/shared-object-version/
https://autotools.io/libtool/version.html

SoapGentoo (approved):

configure.ac
```diff
@@ -85,12 +85,24 @@ dnl      age.
...
+SHARED_VERSION_INFO="$LT_CURRENT:$LT_REVISION:$LT_AGE"
+
+dnl src/version-metadata.rc values.
+dnl WIN_RC_FILEVERSION and WIN_RC_FILEVERSION_STR are calculated from
+dnl LT_XXX variables as (current-age.age.revision).
+dnl WIN_RC_PRODUCTVERSION and WIN_RC_PRODUCTVERSION_STR are calculated from
+dnl PACKAGE_VERSION.
+
+WIN_RC_FILEVERSION="1,0,30"
+WIN_RC_FILEVERSION_STR="1.0.30"
```
evpobr, line 103: @SoapGentoo , i guess it could be calculated, could you help?

SoapGentoo: very hard, because you'd have to compute "backwards-incompatible changes". Let me think about this a bit, for the time being, let's follow the autotools mythbusters instructions.

umlaeute:
hmm, not sure what you mean here.

both `API` and `ABI` refer to the interface between the library and a consumer (dependant).
in the case of libsndfile, they are practically identical.

> If we add MP3 support to the next release, the API version will be 1.1.0 (API compatible changes), and the ABI will be 1.0.31 (without functional changes). More info here.

this is not how semver typically works.
semantic versioning is only about the interface exposed by the library.
i fail to see how adding MP3-support would do anything to the API/ABI (though of course i don't know how you would implement it).

semver can basically be brought down to:
- if you remove any symbols, increase the major number
- if you add any symbols, increase the minor number
- if you neither add nor remove symbols, increase the micro number

adding MP3-support would therefore bump the micro number.
i'm not sure why you think it is somehow related to the API (which boils down to the headers).
to me it seems that you want to have a "feature version" (that is: a versioning scheme that somehow indicates functionality changes of the library, as opposed to it's interface)

however, i'm pretty sure that introducing two different & competing version schemata will only confuse people.
"feature versions" are really common in commercial software, but that is driven by marketing (if the new version costs 1000€, then who would upgrade from 1.0.29 to 1.0.30 (that's about 10000.0.0€/versionnumber!)? but if the upgrade takes you from 98 to 2000, you only pay about 52¢/versionnumber, which is really a bargain!)
semver is a way to overcome "feature versions" in favour of a versioning schema that actually conveys information (rather than marketing blurb).

given that libsndfile is free software, i also see little need to try convince people to buy the latest and greatest.

SoapGentoo:
Again, libtool versions the ABI, and in multiple links it shows how these things are completely uncoupled. You can break API in patch versions (not that I would, but not unheard of either, :wave: `libpython.so`!).

umlaeute:
> Again, libtool versions the ABI, and in multiple links it shows how these things are completely unrelated.

i don't know of which "multiple links" you are talking about.

i also don't know what exactly you mean with "these things".
one of them is obviously the semver ABI-versioning of libtool; but what's the other?

if the *other* is some "feature versioning", then yes: of course they can be different.
but the question is: *why* do you want to do that?

i can only think of two implications:
1. people get noticed that something "big" has changed if the version makes a big leap (like added mp3-support; which - due to the robust interface of libsndfile - would only warrant a bump in the micro-version of the ABI version)
2. people would get confused because now there are two independent versions describing the same release.

i argue that the first point is moot, as we are not trying to *sell* anything; so describing the new feature-set in the documentation should be enough.

> As a result, it is difficult to determine when we added a new API. I suggest separating the API version from the ABI version.

assuming that with "API" you (that is: @evpobr ) mean "backend": how about putting useful information into [ChangeLog](https://github.com/libsndfile/libsndfile/blob/b706e62ead1cb74083c1e7a4db1e3b411aca4a90/ChangeLog)?

SoapGentoo:
https://github.com/libsndfile/libsndfile/pull/652#discussion_r519392538

No-one actually pays attention to the version string in the installed symlink, hence having divergent versions for the API and ABI is a non-issue, like it is for most other libtool using packages. OTOH, Debian having to rebuild their entire userland because upstreams don't know how to version their ABI _is_ a problem, and rightfully so.

evpobr:
> this is not how semver typically works.
semantic versioning is only about the interface exposed by the library.
i fail to see how adding MP3-support would do anything to the API/ABI (though of course i don't know how you would implement it).

Only about *programming* interface. SemVer is about API. Open SemVer specification and try to find ABI word.

> people get noticed that something "big" has changed if the version makes a big leap (like added mp3-support; which - due to the robust interface of libsndfile - would only warrant a bump in the micro-version of the ABI version)

Robust *application binary interface* of libsndfile.

> both API and ABI refer to the interface between the library and a consumer (dependant).

You don't understand difference between API and ABI. You, as programmer, need to know the API version only. This is the version in NEWS, in sndfile.pc etc.

ABI version is for OS. It describes changes in exported function list. That's all. By this version OS knows, that libsndfile.so  1.0.31 is compatible with 1.0.0 and it can be loaded instead of it. But libsndfile 1.0.31 cannot be load instead of 1.1 - minor version bump tells about some new exported functions added, missing in any previous release. ABI of libsndfile is stable - we don't add new public functions for years.

But API version reflects more changes - all the functions, struct, enums, defines in **public header**. And that's why API version will change faster.

For libsndfile we can add new formats and commands - they are enum values and it means that for ABI it is just a micro bump (no changes in exported functions list), but for API  it is minor bump (new backward compatible functionality).

And here we have a problem - libsndfile has only the ABI version, which is also the API version. We can't describe the changes in the API separately. We've added new functionality, but it looks like we've released a patch.

And that's what my pull request is about. I suggest separating the API version from the ABI version, which will allow semantic versioning for the API. The programmer will understand that version 1.1 includes new functionality, and 1.1.1 is just a patch release. In parallel, the ABI version will be updated according to its own rules. But no one will be particularly interested in it.

evpobr:
> this is not how semver typically works.
semantic versioning is only about the interface exposed by the library.
i fail to see how adding MP3-support would do anything to the API/ABI (though of course i don't know how you would implement it).

Only about *programming* interface. SemVer is about API. Open SemVer specification and try to find ABI word.

> people get noticed that something "big" has changed if the version makes a big leap (like added mp3-support; which - due to the robust interface of libsndfile - would only warrant a bump in the micro-version of the ABI version)

Robust *application binary interface* of libsndfile.

> both API and ABI refer to the interface between the library and a consumer (dependant).

You don't understand difference between API and ABI. You, as programmer, need to know the API version only. This is the version in NEWS, in sndfile.pc etc.

ABI version is for OS. It describes changes in exported function list. That's all. By this version OS knows, that libsndfile.so  1.0.31 is compatible with 1.0.0 and it can be loaded instead of it. But libsndfile 1.0.31 cannot be load instead of 1.1 - minor version bump tells about some new exported functions added, missing in any previous release. ABI of libsndfile is stable - we don't add new public functions for years.

But API version reflects more changes - all the functions, struct, enums, defines in **public header**. And that's why API version will change faster.

For libsndfile we can add new formats and commands - they are enum values and it means that for ABI it is just a micro bump (no changes in exported functions list), but for API  it is minor bump (new backward compatible functionality).

And here we have a problem - libsndfile has only the ABI version, which is also the API version. We can't describe the changes in the API separately. We've added new functionality, but it looks like we've released a patch.

And that's what my pull request is about. I suggest separating the API version from the ABI version, which will allow semantic versioning for the API. The programmer will understand that version 1.1 includes new functionality, and 1.1.1 is just a patch release. In parallel, the ABI version will be updated according to its own rules. But no one will be particularly interested in it.

umlaeute:
> And here we have a problem - libsndfile has only the ABI version, which is also the API version. We can't describe the changes in the API separately. We've added new functionality, but it looks like we've released a patch.

i still don't understand which problem you are actually trying to solve?
why do you need to describe the changes in the API separately?
what is the issue if it "looks like we've released a patch"?


> No-one actually pays attention to the version string in the installed symlink, hence having divergent versions for the API and ABI is a non-issue, like it is for most other libtool using packages. OTOH, Debian having to rebuild their entire userland because upstreams don't know how to version their ABI is a problem, and rightfully so.

totally. that's why ABI versioning is really important. ;-)

PR #654, opened 2020-11-12
Exclude repository-configuration from git-archive

umlaeute:
having a .gitignore file in the git-archives (e.g. like the automated
archives provided by github) is rather annoying if the archive gets
imported in other repositories.

my use-case is the Debian packaging, where we really want to see any files created/modified during the build process.


evpobr:
Ok, thank you @umlaeute .

PR #659, opened 2020-11-20
configure: Use version-script when compiling with clang on unix

tstellar:
This fixes incompatibilities with the clang and gcc builds, and also
matches what is done in the CMake builds.


SoapGentoo (approved):

evpobr:
Thanks @tstellar

Issue #643, opened 2020-10-22
missing samples when doing a partial read of ogg file from index till the end of file


krisfed:
Hi all,

We are in the process of upgrading from 1.0.28 to 1.0.30, and we noticed this issue in the new release.

Looks like it happens when you seek to a start index that is >=88201 before reading. The read data seems to match the data at end of the file, but some samples right after the start index are missing. The number of missing samples seems to always be a power of two.

Here is simple reproduction code with an example:

```
#include <iostream>
#include "sndfile.h"

int main()
{
	// open file
       SF_INFO fileInfo;

       SNDFILE* file_sf = sf_open("mytest.ogg", SFM_READ, &fileInfo);
	
	// seek
	sf_count_t startSample = 389999;
	sf_count_t framesSeeked = sf_seek(file_sf, startSample, SEEK_SET);

        std::cout << "startSample: " << startSample << std::endl;
	std::cout << "framesSeeked: " << framesSeeked << std::endl;
	
	// read
	sf_count_t framesRead = 0;
	
	size_t frameSize = 65536;
	size_t sizeInBytes = 4 * frameSize * 2;// type size * num samples * num channels
	uint8_t* buffer = new uint8_t[sizeInBytes];
	
	framesRead = sf_readf_float(file_sf, reinterpret_cast<float*>(buffer), frameSize);
	
        std::cout << "frameSize: " << frameSize << std::endl;
	std::cout << "sizeInBytes: " << sizeInBytes << std::endl;
	std::cout << "framesRead: " << framesRead << std::endl;
	
	sf_close(file_sf);

    return 0;

}
```

For this example, the total number of samples is 396519. If we start at 389999 and try to read a framesize (65536) we should get all the remaining samples in the file - 6520. But with 1.0.30 release I only get 5496 (missing 1024 samples).

Here is the output of the above repro code for 1.0.28:
```
startSample: 389999
framesSeeked: 389999
frameSize: 65536
sizeInBytes: 524288
framesRead: 6520
```
And for 1.0.30:
```
startSample: 389999
framesSeeked: 389999
frameSize: 65536
sizeInBytes: 524288
framesRead: 5496
```
(both ran on macOS 10.15.3, but we see the issue on Windows too)

The file used in the example is attached - [mytest.zip](https://github.com/libsndfile/libsndfile/files/5425485/mytest.zip). It was randomly generated, but I see this issue on other .ogg files that are long enough. Here is the snd-info output:
```
[...]

Issue #194, opened 2017-01-30
pcm overflows for ogg files


derselbst:
I came across an unexpected behaviour when decoding loud ogg files. If a sample seems to have max value, it will overflow when decoding it with sndfile's integer read functions. The image shows the waveform of an [example ogg file](https://raw.githubusercontent.com/SuperTux/data/1bbcf668304466b37868356e0d0278f2132c90d2/music/fortress.ogg) directly imported to audacity. The waveform right below shows the same ogg file decoded to raw `.wav` before import.
![sndfile_bug](https://cloud.githubusercontent.com/assets/8152480/22437015/9c072c12-e726-11e6-9f2e-6b37fcc800bb.png)

To reproduce:
* grab the [example file](https://raw.githubusercontent.com/SuperTux/data/1bbcf668304466b37868356e0d0278f2132c90d2/music/fortress.ogg)
* play it back using `sndfile-play`
* on linux, everything should sound fine so far, since `sndfile-play` uses `sf_read_float()` here
* adjust `sndfile-play` to make it decoding using  `sf_read_int()` (also occurs with `sf_read_short()`)
* when listening again you should hear the first crack after about 8 seconds

To avoid the overflow I have to run `SFC_SET_SCALE_FLOAT_INT_READ`, however (from my understanding) [documentation](http://www.mega-nerd.com/libsndfile/api.html#note2) only suggests this when having `SF_FORMAT_FLOAT` or `SF_FORMAT_DOUBLE`.

So if it is intended to run this command also for `SF_FORMAT_VORBIS`, an update of docs might be fine? Or am I missing something?

erikd:
Yes, I think the documentation should be updated. Will keep this ticket open until I do that.

kcat:
Having the option to clamp the integer samples instead of wrapping would be good, too. The distortion from clamped samples is far less noticeable and less disturbing compared to wrapped samples, but having to attenuate the whole file to avoid a few out of range samples seems like overkill.

Also something else I noticed relating to pcm overflows, as mentioned in kcat/openal-soft#505, reading ogg samples as int scales like `pcm [n][j] * 2147483647.0f`. But single-precision floats can't represent 2147483647, so it gets rounded at compile-time to the closest representable value, `2147483648.0f`. When +1.0 is scaled and converted to int, 2147483648 wraps to -2147483648, causing a file that even stays within the -1...+1 range to overflow sample values on int reads.

derselbst:
[...]

PR #663, opened 2020-12-07
Implement fast SSE2 lrint and lrintf

evpobr:
`lrint`() and `lrintf`() are ridiculously slow when libsndfile is built using Visual C++.

On x86 and AMD64 platforms this patch adds SSE2 optimized versions.

It is safe to assume that non-SSE2 processors are no longer available to avoid unnecessary checks.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,7 @@
 Unreleased
+  * Implement fast SSE2 optimized psf_lrintf() and psf_lrintf() functions to
+    improve perfomance when libsndfile is built using Visual C++ (especially)
+    and other compilers on x86 and AMD64 platforms.
   * Documentation:
     * Move site to new URL: http://libsndfile.github.io/libsndfile/
     * Convert documentation pages from HTML to Markdown
--- a/cmake/SndFileChecks.cmake
+++ b/cmake/SndFileChecks.cmake
@@ -67,6 +67,7 @@ check_include_file (stdint.h		HAVE_STDINT_H)
 check_include_file (sys/time.h		HAVE_SYS_TIME_H)
 check_include_file (sys/types.h		HAVE_SYS_TYPES_H)
 check_include_file (unistd.h		HAVE_UNISTD_H)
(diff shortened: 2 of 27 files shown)


evpobr:
The tests are here: https://quick-bench.com/q/OabKT-gEOZ8CYDriy1JEwq1lEsg. Note this is GCC, Visual C++ will show much more impressive results :smile: .

evpobr:
@cmeister2 , OSSFuzz tests failling?

cmeister2:
Expect that the python version in use is Python 2 and that the main repository is now using Python 3-only code. Do you want me to raise a separate PR to fix?

evpobr:
Yes :smile:

Be-ing:
I think this broke Mixxx: https://github.com/mixxxdj/mixxx/pull/3615#issuecomment-775605582

evpobr:
@Be-ing , need more info. We pass all our tests. Could it be related to #672?

evpobr:
The code is taken from [Agner Fog's Optimizing C++ manual](https://www.agner.org/optimize/optimizing_cpp.pdf). I don't think it is incorrect.

Be-ing:
I tried reverting this and the other SSE related commits and it did not fix Mixxx's tests. I am bisecting now..

Be-ing:
> @Be-ing , need more info. We pass all our tests. Could it be related to #672?

Seems plausible, but why would it only be an issue for Mixxx now with 1.0.31 if #194 has been present for years? Perhaps #680?

Be-ing:
Ah, we have a workaround for #643 in Mixxx's SoundSourceSndfile class that denylists libsndfile 1.0.30 as a decoder for ogg files. I guess we need to extend that to versions >= 1.0.30 until #643 is fixed. Sorry for the noise here.

evpobr:
Don't worry, thanks for report.

arthurt:
The SSE related commits could not effect the sample-count issue.

PR #668, opened 2020-12-10
Make SSE2 configurable for CMake builds

evpobr:


evpobr:
@SoapGentoo , can you help me with this? We need to add `--enable-see2` option to Autotools:

1. Detect x86 or x64 host CPU
2.  Check `immintrin.h`
3. Add `--enable-sse2` option for x86 and x64 CPUs with `immintrin.h` to control next steps 
3. If `--enable-sse2` define USE_SSE2 in `config.h`
4. If `--enable-sse2`  and x86 CPU check `-msse2` compiler and add it needed.

I've found useful code in FLAC's [`configure.ac`](https://github.com/xiph/flac/blob/master/configure.ac#L110).

I have a rough idea of how to do the individual steps, but I can't link them all together.

SoapGentoo:
@evpobr I'll help you on the weekend. Here's my take though: we should just test whether `immintrin.h` works when `--enable-sse2` is enabled, and bomb out if it doesn't work. We shouldn't silently disable it if it doesn't work.

evpobr:
To be honest, I'm not entirely sure I did it right. Now `ENABLE_SSE2` enables/disables both the -msse2 flag and all SSE2 code. It may be more correct to check only the need of `-msse2` flag and the `ENABLE_SSE2` option should only control this. The code itself at the source level should check the compiler defines and HAVE_EMMINTRIN_H, something like this:

```c
#if (defined (__SSE2__) || defined (_M_AMD64) || (defined (_M_IX86_FP) && (_M_IX86_FP >= 2)) && HAVE_IMMINTRIN_H)
#define USE_SSE2
#endif
```

in this case, you can allow the user to disable the-msse2 flag and add, say, - mavx2.

PR #680, opened 2020-12-31
Add .ogg to major_formats

makitsune:
I wrote this bit of code for getting all the available extensions supported by libsndfile:

```cpp
QList<QString> extensions;

int count;
sf_command(nullptr, SFC_GET_FORMAT_MAJOR_COUNT, &count, sizeof(int));

for (int i = 0; i < count; i++) {
	SF_FORMAT_INFO info;
	info.format = i;
	sf_command(nullptr, SFC_GET_FORMAT_MAJOR, &info, sizeof(info));
	extensions.push_back(QString(info.extension).toLower());
};

return extensions;
```

Eventually I find out that my `ogg` files are ignored because the extension isn't included in `major_formats`

Perhaps we can also add `ogv` if that's supported. All extensions `oga`, `ogg` and `ogv` hold vorbis audio data.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -71,7 +71,7 @@ static SF_FORMAT_INFO const simple_formats [] =
 		},
 
 	{	SF_FORMAT_OGG | SF_FORMAT_VORBIS,
-		"Ogg Vorbis (Xiph Foundation)", "oga"
+		"Ogg Vorbis (Xiph Foundation)", "ogg"
 		},
 #endif
 


makitsune:
Seems the tests fail because the description or enum is the same. I'm not sure what to do here.

On the Wikipedia page for ogg: https://en.wikipedia.org/wiki/Ogg
It states: "Before 2007, the .ogg filename extension was used for all files whose content used the Ogg container format. Since 2007, the Xiph.Org Foundation recommends that .ogg only be used for Ogg Vorbis audio files."

evpobr (commented):
Tests are failing.

evpobr:
> I wrote this bit of code for getting all the available extensions supported by libsndfile:

It seems like `SFC_GET_FORMAT_MAJOR` doesn't support multiple extensions. Probably we need a new command.

makitsune:
We could also just rename oga to ogg since Xiph recommends thats

evpobr:
Hmm, @erikd, @SoapGentoo , @arthurt , what do you think?

arthurt:
I don't think this is the correct way to go about this.

Firstly, I don't think it makes sense to add multiple entries for each major format.

Second, SFC_GET_FORMAT_MAJOR_COUNT returns info about major formats, of which the Ogg container is one.

Ogg as a container is different from other containers like AVI and MKV, being mostly known and associated with the one codec: Vorbis. So much so that Xiph.org, the creator of both, concedes now that `.ogg` implies Vorbis-in-Ogg. *1

Ogg is also used as a container for Opus streams. As per the OggOpus spec, an Ogg file with one Opus stream should have a `.opus` extension. Likewise, Speex is contained in Ogg, and has the extension of `.spx`.

So, what is the answer to the question "What is the file extension of an Ogg container containing an audio stream" without knowing what the codec (minor format) is? Can't use `.ogg` as that implies Vorbis. Can't use `.opus` or `.spx` either obviously.

This is where `.oga` comes in. `.oga` means "some audio stream contained in Ogg," which is the correct meaning here.

---

One thing that should be updated however is changing the extension for OggVorbis in SFC_GET_SIMPLE_FORMAT: 

```
command.c:73
---
	{	SF_FORMAT_OGG | SF_FORMAT_VORBIS,
		"Ogg Vorbis (Xiph Foundation)", "oga"
		},
```

This should arguable be changed to `.ogg` now.

This is near the same location that defines `.opus` as `SF_FORMAT_OGG | SF_FORMAT_OPUS`

evpobr:
Oops, i've edited @arthurt comment, sorry.

It should be edited here:

https://github.com/libsndfile/libsndfile/blob/9113112445eaddac3d5e14332151c50e73105184/src/command.c#L73-L75

makitsune:
I've renamed oga to ogg in `simple_formats`. However, I need a reliable way to know which file formats are supported.

I think I should probably use `format_map` in `programs/common.c` but there's only 2 functions that access it:
- `sfe_file_type_of_ext` filename to format 
- `sfe_dump_format_map` just prints `format_map`

But there's no function to fetch all the extensions one by one like I was doing with my code snippet earlier. Though I just realized I can't get to `programs/common.c` from my program, so that's not going to work either.

Just as a side note, it looks like `aiff` is missing from `format_map` as well. Only `aif` is present and both are identical according to https://en.wikipedia.org/wiki/Audio_Interchange_File_Format

evpobr:
We need to implement a new extend format info. Planned for future.

Be-ing:
I think this has made #194 an issue with 1.0.31 for software wasn't affected by #194 before.

PR #681, opened 2021-01-03
ID3v2 skip fixes

arthurt:
Fix a bug in id3.c where large ID3v2 headers would fail to seek to the embedded file. ID3v2 tags are skipped over, treating the wrapped file as an embedded file. `id3_skip()` sets the embedded file offset to be just after the ID3v2 tag ends, then seeks the `psf_binheader()` to the fileoffset.

`psf_binheader_read()` accomplishes seeks by either reading and buffering bytes until the offset is reached, or if the buffer would be too large, seeking the underlying file instead using `psf_fseek()`. `psf_fread()` does not take the embedded file offset into account, while `psf_fseek()` does.

In the case of ID3 tags larger than the binheader buffer max, this has the effect of seeking to twice the id3v2 tag length, as
`psf_fseek(fileoffset, SEEK_CUR)` adds the fileoffset to the offset argument again. This lands in the middle of a stream which then cannot be identified.

To resolve the issue, seek the binheader *before* setting the embedded file offset.

Also, identify ID3v2.2 and ID3v2.4 headers. Previously only ID3v2.3 headers were identified.

An as-yet unresolved issue is that `id3_skip()` of large ID3v2 tags does not work on pipes as pipes do not support seeking.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -877,6 +877,7 @@ header_seek (SF_PRIVATE *psf, sf_count_t position, int whence)
 				psf_bump_header_allocation (psf, position) ;
 			if (position > psf->header.len)
 			{	/* Too much header to cache so just seek instead. */
+				psf->header.indx = psf->header.end ;
 				psf_fseek (psf, position, whence) ;
 				return ;
 				} ;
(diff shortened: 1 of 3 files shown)


evpobr:
@arthurt , if you think it is ready just merge it.

PR #690, opened 2021-01-20
fix api.md table error

zodf0055980:
Correct md file table.


evpobr:
@zodf0055980 , thanks.

PR #693, opened 2021-01-24
Release 1.0.31

SoapGentoo:


evpobr:
Have you checked library SONAME on Linux platfrom?

SoapGentoo:
`1.0.30`:
```
/usr/lib64/libsndfile.so -> libsndfile.so.1.0.30
/usr/lib64/libsndfile.so.1 -> libsndfile.so.1.0.30
/usr/lib64/libsndfile.so.1.0.30
```
(i.e. SONAME = `libsndfile.so.1`)

this tentative release (`1.0.31`):
```
libsndfile.so -> libsndfile.so.1.0.31
libsndfile.so.1 -> libsndfile.so.1.0.31
libsndfile.so.1.0.31
```
(i.e. still SONAME = `libsndfile.so.1`)
the SONAME hasn't changed, but the versioned filename has, so everything is fine :+1:

evpobr:
Could you rename commit message to "Release 1.0.30" as i did in my releases? Traditions are good.  And it is not preparation, this commit actually makes new release.

SoapGentoo:
you mean "Release 1.0.31"?

evpobr:
README file (not README.md) is not updated since my release, version still 1.0.29. My fault. Better to fix it.

evpobr:
> you mean "Release 1.0.31"?

Yep.

SoapGentoo:
@evpobr done, are we okay here? If so, I will just push this commit straight to master

evpobr:
CMake still uses `${PROJECT_VERSION}` and it is API version:

https://github.com/libsndfile/libsndfile/blob/942ae829be5c51b080c7181902ca6b547b7bfb9e/CMakeLists.txt#L383-L386

SoapGentoo:
`VERSION ${PROJECT_VERSION}` in CMake is fine, it just affects the filename, *not* the SONAME. Should we change `SOVERSION ${PROJECT_VERSION_MAJOR}` to `SOVERSION 1` to bring it in line with libtool?

evpobr:
> `VERSION ${PROJECT_VERSION}` in CMake is fine, it just affects the filename, _not_ the SONAME. Should we change `SOVERSION ${PROJECT_VERSION_MAJOR}` to `SOVERSION 1` to bring it in line with libtool?

In the code mentioned above it sets SOVERSION. But format is SemVer unlike libtool versioning:

```cmake
 set_target_properties (sndfile PROPERTIES 
 	SOVERSION 1
 	VERSION 1.0.31
 	) 
``cmake

For now it works, but when API and ABI will diverge, we will get into trouble.

SoapGentoo:
as I mentioned in the libsamplerate change, if we have an ABI break without an API break (say, changing function arguments from `int` to `long long`) the SOVERSION and major version will necessarily diverge, hence why I'm against automatic ABI versioning. My suggestion is to keep `VERSION` automatic, but `SOVERSION` manually managed.

zvezdochiot:
@evpobr say:
>For now it works, but when API and ABI will diverge, we will get into trouble.

See:
https://github.com/mozilla/mozjpeg/blob/44f3f8a544e46f1859d45177bb0c7f268ea45c29/CMakeLists.txt#L8-L12

evpobr:
We can make it in next release, this is not showstopper.

> as I mentioned in the libsamplerate change, if we have an ABI break without an API break (say, changing function arguments from `int` to `long long`) the SOVERSION and major version will necessarily diverge, hence why I'm against automatic ABI versioning. My suggestion is to keep `VERSION` automatic, but `SOVERSION` manually managed.

But we got an issue with the `0.2.0` libsamplerate release due to such assumptions.

evpobr:
> @evpobr say:
> 
> > For now it works, but when API and ABI will diverge, we will get into trouble.
> 
> See:
> https://github.com/mozilla/mozjpeg/blob/44f3f8a544e46f1859d45177bb0c7f268ea45c29/CMakeLists.txt#L8-L12

They are restricted to 2.8.12, and that version has no `${PROJECT_VERSION_XXX}` variables set in `project()`.

SoapGentoo:
yes, because in libsamplerate, the libtool filename was directly computed from `major.minor.patch` (which is exactly what libtool recommends against). The real problem here is that libtool and CMake version ABI differently using the given tuple, and libtool's algorithm to compute the filename (and also `SONAME`) is arcane and not intuitively obvious. In libtool, you never specify the `SOVERSION` explicitly, instead you calculate it from `current` and `age`, whereas in CMake, you specify the SOVERSION directly.

Hence my suggestion: let's specify `SOVERSION` manually in CMake, and then we bump it if we ever run into an ABI break. Computing the filename on the basis of `${PROJECT_VERSION}`, as that is guaranteed to be monotonically increasing.

evpobr:
Ok, don't touch it. We will think about it in next release.

evpobr:
But do touch NEWS, version and date are unspecified.

evpobr:
And **History** section in `docs/index.md`.

And then we are ready.

SoapGentoo:
@evpobr fine now?

evpobr:
Ok.

evpobr (approved):

evpobr:
Ok, thank you!

PR #704, opened 2021-02-13
Change test file names to avoid conflicts when running tests

bobsayshilol:
`ctest -j <N>` can fail sporadically due to tests sharing the same temporary file names. This PR fixes the issue by changing the names of all those that conflict.

I'm not sure if `get_unique_test_name()` is the best approach, but it was simpler than changing every string.

### DIFF
--- a/tests/compression_size_test.c
+++ b/tests/compression_size_test.c
@@ -37,6 +37,8 @@
 #define	SAMPLE_RATE		16000
 #define	DATA_LENGTH		(SAMPLE_RATE)
 
+static const char CMP_TEST_PREFIX[] = "cmp" ;
+
 static float data_out [DATA_LENGTH] ;
 
 static inline float
@@ -52,6 +54,8 @@ vorbis_test (void)
 	SF_INFO sfinfo ;
 	float max_abs = 0.0 ;
 	unsigned k ;
+
+	get_unique_test_name (&filename, CMP_TEST_PREFIX) ;
 
 	print_test_name ("vorbis_test", filename) ;
 
@@ -122,6 +126,8 @@ compression_size_test (int format, const char * filename)
 	int q3_size, q6_size ;
 	double quality ;
 	int k ;
+
+	get_unique_test_name (&filename, CMP_TEST_PREFIX) ;
 
 	snprintf (q3_fname, sizeof (q3_fname), "q3_%s", filename) ;
 	snprintf (q6_fname, sizeof (q6_fname), "q6_%s", filename) ;
--- a/tests/utils.tpl
+++ b/tests/utils.tpl
@@ -104,6 +104,8 @@ void	count_open_files (void) ;
 void	increment_open_file_count (void) ;
 void	check_open_file_count_or_die (int lineno) ;
 
+void	get_unique_test_name (const char ** filename, const char * test) ;
+
 #ifdef SNDFILE_H
 
 static inline void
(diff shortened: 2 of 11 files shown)


evpobr:
Autotools build fails:

```
ERROR: files left in build directory after distclean:
./lct_opus.opus
make[1]: *** [distcleancheck] Error 1
Makefile:3429: recipe for target 'distcleancheck' failed
make[1]: Leaving directory '/home/runner/work/libsndfile/libsndfile/libsndfile-1.0.31/_build/sub'
Makefile:3356: recipe for target 'distcheck' failed
make: *** [distcheck] Error 1
Error: Process completed with exit code 2.
```

bobsayshilol:
> Autotools build fails:

Yeah, and looking through the logs I see `opus.opus` appear more than once which my script didn't catch from a test run since I wasn't building with that enabled locally, so more fixes are required there too.

I had a brief look at the autotools issue but I don't know enough about autotools to pinpoint where it decides what to cleanup, so can you point me in the right direction please?

evpobr:
For Autotools ask @SoapGentoo :smile:

SoapGentoo:
@bobsayshilol I'll have a look

SoapGentoo:
@bobsayshilol could you squash and rebase on latest `master`? I can't reproduce the CI issue locally

bobsayshilol:
> @bobsayshilol could you squash and rebase on latest `master`? I can't reproduce the CI issue locally

I've fixed up the remaining filename collisions, and then rebased and squashed onto master but it looks like it's still failing:
```
 ERROR: files left in build directory after distclean:
./lct_opus.opus
make[1]: *** [distcleancheck] Error 1
```

It's strange that `macos-autotools` is passing though since presumably it's also using automake?

SoapGentoo:
@bobsayshilol can I push to your branch? I have a hunch

bobsayshilol:
Sure, go for it.

tests/utils.tpl
```diff
@@ -851,6 +853,15 @@ check_open_file_count_or_die (int lineno)
 #endif
 } /* check_open_file_count_or_die */
 
+void
+get_unique_test_name (const char ** filename, const char * test)
+{	static char	buffer [1024] ;
+
+	snprintf (buffer, sizeof (buffer) - 1, "%s_%s", test, *filename) ;
```
SoapGentoo, line 860: @bobsayshilol foudn the bug. For the final review, could you amend this? `snprintf` assumes the maximum length it can use is `n-1`, i.e. it's already `NUL`-termination aware and doesn't require passing `n-1` into it:

```suggestion
	snprintf (buffer, sizeof (buffer), "%s_%s", test, *filename) ;
```

bobsayshilol:
Nice catch @SoapGentoo!

tests/compression_size_test.c
```diff
@@ -37,6 +37,8 @@
 #define	SAMPLE_RATE		16000
 #define	DATA_LENGTH		(SAMPLE_RATE)
 
+#define	CMP_TEST_PREFIX	"cmp"
```
SoapGentoo, line 40: Let's move away from macros, which are ugly. Let's go with [Ulrich Drepper's suggestion](https://udrepper.livejournal.com/15119.html):
```suggestion
static const char CMP_TEST_PREFIX[] = "cmp";
```

SoapGentoo (changes requested):

tests/utils.tpl
bobsayshilol: Derp. Good spot!

tests/compression_size_test.c
bobsayshilol: Ah that was a leftover from my original approach of prepending `*_TEST_PREFIX` to every filename, which quickly got out of hand. I'll change them over.

SoapGentoo (approved):
Thanks, I'm happy now, @evpobr please check too

evpobr (approved):

PR #705, opened 2021-02-14
Add commit message requirements to CONTRIBUTING.md

evpobr:
1. Separate subject from body with a blank line
2. Limit the subject line to 50 characters
3. Capitalize the subject line
3. Do not end the subject line with a period
4. Use the imperative mood in the subject line
5. Wrap the body at 72 characters
6. Use the body to explain what and why vs. how

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,7 @@
 * Patches should always be submitted via a either Github "pull request" or a
   via emailed patches created using "git format-patch".
 * Patches for new features should include tests and documentation.
+* Commit messages should follow the ["How to Write a Git Commit Message"](https://chris.beams.io/posts/git-commit/) guide.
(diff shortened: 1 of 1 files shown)


evpobr:
> I agree with it. I would add, I do like the practice of if a commit deals with one feature or source prefixing the commit with the name of it followed by a colon. eg the ogg_vorbis: Change blah or ogg: Do blah. Thoughts?

@arthurt:

I like the clean version with no prefixes. I suggest adding a note to the `CONTRIBUTING.md` that the commit message may contain prefix.

@SoapGentoo ?

SoapGentoo:
I'm fine either way. Gentoo uses a prefix, other projects don't. In my corporate codebase, I tend to use prefixes, because they make things slightly more comprehensible, but I don't mind either way.

PR #709, opened 2021-02-15
Fix for Issue 643: Vorbis Seek Offset.

### LINKED ISSUE #643: missing samples when doing a partial read of ogg file from index till the end of file

arthurt:
Previously the `vorbis_seek()` function did not take the lapped nature of Vorbis block into account when using the bisection search.

Also, introduces `vorbis_skip()`, which can skip through packets without having to decode all the way to samples.

Fixes #643

### DIFF
--- a/src/ogg_vorbis.c
+++ b/src/ogg_vorbis.c
@@ -1,4 +1,5 @@
 /*
+** Copyright (C) 2018-2021 Arthur Taylor <art@ified.ca>
 ** Copyright (C) 2002-2016 Erik de Castro Lopo <erikd@mega-nerd.com>
 ** Copyright (C) 2002-2005 Michael Smith <msmith@xiph.org>
 ** Copyright (C) 2007 John ffitch
@@ -19,7 +20,7 @@
 */
 
 /*
-**  Much of this code is based on the examples in libvorbis from the
+** Much of this code is based on the examples in libvorbis from the
 ** XIPHOPHORUS Company http://www.xiph.org/ which has a BSD-style Licence
 ** Copyright (c) 2002, Xiph.org Foundation
 **
@@ -78,6 +79,9 @@
 
 #include "ogg.h"
 
+/* How many seconds in the future to not bother bisection searching for. */
+#define VORBIS_SEEK_THRESHOLD 2
+
 typedef int convert_func (SF_PRIVATE *psf, int, void *, int, int, float **) ;
 
 static int	vorbis_read_header (SF_PRIVATE *psf) ;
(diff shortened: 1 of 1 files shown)


evpobr:
Great, you took the time, thanks. You have done all the work, so it will be fair if you moderate the changes.

PS Did you see #705? What do you think?

arthurt:
> PS Did you see #705? What do you think?

I agree with it. I would add, I do like the practice of if a commit deals with one feature or source prefixing the commit with the name of it followed by a colon. eg the `ogg_vorbis: Change blah` or `ogg: Do blah`. Thoughts?

evpobr:
Let's continue the discussion in #705.

evpobr (approved):

Be-ing:
Mixxx tests [still fail](https://github.com/Be-ing/mixxx/runs/1908562676?check_suite_focus=true#step:19:6434)

PR #713, opened 2021-02-20
oss-fuzz ASAN fixes

bobsayshilol:
This PR fixes the security issues spotted by oss-fuzz in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26026 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26803. Of the other publicly visible security issues only https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27503 looks frightening, and that looks like it'll go away once the UBSan issues in ALAC are fixed.


evpobr:
Thanks again @bobsayshilol !

PR #715, opened 2021-02-28
Revert "Fix zero-as-null-pointer-constant-warning"

brianlheim:
This reverts commit 2269af84173c1f70b6d700fa167e1e8ea01d7247.

The code which defined nullptr mysteriously broke the Windows build for
SuperCollider. After some investigation I discovered this commit to be the
cause, as MSVC by default does not set __cplusplus to anything greater
than 199711L unless explicitly asked to.

In general, #defining a keyword like nullptr in a library header is not
courteous, as it can affect compilation of later code and cause
significant confusion. This was exactly the situation I encountered, and
it took several days to track down the source.

The stated reason for the commit reverted by this was to prevent a
specific warning from appearing during compilation. However, there are
a few less intrusive ways of accomplishing this. For example, by using
`#pragma`s offered by all major compilers which temporarily disable
warnings, or by including headers with `-isystem` or any similar
mechanism which disables warnings from "system" headers. For code which
is intended to support C++98 and newer standards, preventing all
compilers from emitting any warnings is a losing battle. It is easier
in my view for consuming code to handle this as it needs to.

There are several options here for resolving this. I could just fix my
project, but that would not prevent others from encountering the same
issue. I could try for a clever-er version of this using a macro not
named nullptr and an additional check for MSVC, but this seems to offer
diminishing returns considering how easily any warnings about NULL can
be silenced, and carries the risk of introducing new and unforeseen
problems. So, I have here simply reverted the commit so the code uses
NULL once again.

---

I'm happy to discuss alternate ways of solving this. I guess I would first ask whether this header needs to support C++98 at all? It doesn't seem to be stated explicitly.

### DIFF
--- a/include/sndfile.hh
+++ b/include/sndfile.hh
@@ -57,12 +57,6 @@
 #include <string>
 #include <new> // for std::nothrow
 
-#if __cplusplus < 201100
-#ifndef nullptr
-#define nullptr NULL
-#endif
-#endif
-
 class SndfileHandle
 {	private :
 		struct SNDFILE_ref
@@ -78,7 +72,7 @@ class SndfileHandle
 
 	public :
 			/* Default constructor */
-			SndfileHandle (void) : p (nullptr) {} ;
+			SndfileHandle (void) : p (NULL) {} ;
 			SndfileHandle (const char *path, int mode = SFM_READ,
 							int format = 0, int channels = 0, int samplerate = 0) ;
 			SndfileHandle (std::string const & path, int mode = SFM_READ,
@@ -99,9 +93,9 @@ class SndfileHandle
 			SndfileHandle & operator = (const SndfileHandle &rhs) ;
 
 		/* Mainly for debugging/testing. */
-		int refCount (void) const { return (p == nullptr) ? 0 : p->ref ; }
+		int refCount (void) const { return (p == NULL) ? 0 : p->ref ; }
 
-		operator bool () const { return (p != nullptr) ; }
+		operator bool () const { return (p != NULL) ; }
 
 		bool operator == (const SndfileHandle &rhs) const { return (p == rhs.p) ; }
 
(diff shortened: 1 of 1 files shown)


evpobr:
Hi @brianlheim, i think we can remove C++ 11 check, because Visual Studio support `nullptr` starting from 2012 version.

brianlheim:
@evpobr is this header meant to require C++11? i would recommend documenting this if so. currently, i can compile it with gcc in C++98 mode.

evpobr:
Hmm. Well, what if

```c
#if ((defined (_MSC_VER) && (_MSC_VER >= 1600)) || (__cplusplus >= 201100))
#define SF_NULL nullptr_t
#else
#define SF_NULL NULL
#endif
```

brianlheim:
as I said in my commit message I think this is unnecessarily complicated, but I see no immediate issue with that check, assuming that you are 100% sure MSVC 2010 didn't have nullptr. you should probably then document that this header is meant to support C++98 and versions of MSVC prior to 2012. or I can add a comment myself to the top of this file.

evpobr:
Actually, `Zero as null pointer` warning comes from:

```c
if (p != NULL)
```

Why not simply use:

```c
if (p)
```

We will fix your issue and that warning. Everybody is happy. And of course you can add documentation comment.

brianlheim:
NULL is not just used in checks, but also constructor initializer lists. to my knowledge it can't be removed there because brace-initialization is also not available until C++11.

evpobr:
@MartinDelille , how did you get these warnings?

Tried to revert your commit and compile with this `-Wzero-as-null-pointer-constant` flag, no warnings.

brianlheim:
i see it with clang but not gcc. gcc's preprocessor expands `NULL` to an special compiler intrinsic, `__null`, instead of `0`.

evpobr:
@brianlheim , can you add my version check update and add documentation commit?

MartinDelille:
@brianlheim I'm really sorry for the inconvenience... 😞 

Thank you for your explanation I'll be more cautious for my next contribution to C++ projects. Using `#pragma` around include is indeed perfectly fine!

brianlheim:
> can you add my version check update and add documentation commit?

done!

> I'm really sorry for the inconvenience... disappointed
> Thank you for your explanation I'll be more cautious for my next contribution to C++ projects. Using #pragma around include is indeed perfectly fine!

No worries! I understand you were just trying to improve the project. :)

evpobr:
Thanks to all 👍

PR #723, opened 2021-03-25
WIP: Support MPEG Layer 3 in WAV

arthurt:
# Support for MPEGLAYER3WAVEFORMAT.

MPEGLAYER3WAVEFORMAT is a method to embed a MPEG layer III audio stream inside of a Wav file. This is a long-existent format, created sometime around Windows 95. The format allows for VBR streams, and includes important information about the stream for gapless playback that was later solved by Info/Xing header it predates.

Many mp3 files are actually embedded inside of a WAV file. Because of the frame sync and junk tolerance of MPEG decoders, most of such files are suffixed with ".mp3" and will usually work just fine with the MPEG decoder ignoring the RIFF part of the file. Users may not be aware that their mp3 file actually has a wav RIFF header.

This presents a problem for Libsndfile, which first tryies to detect known containers from their data signatures, regardless of suffix. When it sees a WAV RIFF header it will try and parse the file as a WAV file but then fail for lack of support for MPEGLAYER3WAVEFORMAT.

This PR adds support for reading and writing the MPEG_LAYER_III subformat in the WAV container.

## TODO
- [x] MPEGLAYER3 fmt parsing
- [ ]  MPEGLAYER3 fmt writing
  - [ ] Plumb through details from `mpeg_l3_encoder.c`
- [ ] Modify MPEG and WAV Tests
- [ ] Support for Gapless decoding using WAV header metadata instead of Info/Xing header metadata. (Will require coordination with mpg123 upstream.)

### DIFF
--- a/src/mpeg_decode.c
+++ b/src/mpeg_decode.c
@@ -568,7 +568,6 @@ mpeg_decoder_init (SF_PRIVATE *psf)
 	mpg123_param (pmp3d->pmh, MPG123_ADD_FLAGS, MPG123_NO_FRANKENSTEIN, 1.0) ;
 #endif
 
-	psf->dataoffset = 0 ;
 	/*
 	** Need to pass the first MPEG frame to libmpg123, but that frame was read
 	** into psf->binheader in order that we could identify the stream.
@@ -578,8 +577,8 @@ mpeg_decoder_init (SF_PRIVATE *psf)
 		** Can't seek, so setup our libmpg123 io callbacks to read the binheader
 		** buffer first.
 		*/
-		psf_binheader_readf (psf, "p", 0) ;
-		pmp3d->header_remaining = psf_binheader_readf (psf, NULL) ;
+		psf_binheader_readf (psf, "p", psf->dataoffset) ;
+		pmp3d->header_remaining = psf_binheader_readf (psf, NULL) - psf->dataoffset ;
 
 		/* Tell libmpg123 we can't seek the file. */
 		mpg123_param (pmp3d->pmh, MPG123_ADD_FLAGS, MPG123_NO_PEEK_END, 1.0) ;
--- a/src/mpeg_l3_encode.c
+++ b/src/mpeg_l3_encode.c
@@ -496,11 +496,11 @@ mpeg_l3_encode_write_short_stereo (SF_PRIVATE *psf, const short *ptr, sf_count_t
 	MPEG_L3_ENC_PRIVATE *pmpeg = (MPEG_L3_ENC_PRIVATE*) psf->codec_data ;
 	sf_count_t total = 0 ;
 	int nbytes, writecount, writen ;
-	const sf_count_t max_samples = SF_MIN (ARRAY_LEN (ubuf.sbuf), pmpeg->frame_samples) ;
 
 	if ((psf->error = mpeg_l3_encoder_construct (psf)))
 		return 0 ;
 
+	const sf_count_t max_samples = SF_MIN (ARRAY_LEN (ubuf.sbuf), pmpeg->frame_samples) ;
 	while (len)
 	{	writecount = SF_MIN (len, max_samples) ;
 		/*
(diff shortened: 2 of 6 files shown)


arthurt:
@sobukus What would be the feasibility of adding to mpg123 support for adding the ability of a caller to set the codec delay and total stream length?

In a MPEGLAYER3 fmt chunk we get the codec delay and from the FACT chunk the total pre-encoded stream sample count. I don't think streams contained in WAV MPEGLAYER3 should contain a Xing/Info header, but if libsndfile knows and can parse the delay and total length from the WAV headers, it would be useful to be able to tell mpg123 about it.

I see there are the existing flags `MPG123_START_FRAME` and `MPG123_DECODE_FRAMES`, but neither appear to be wired up.

Admittedly, I don't know how many MPEGLAYER3 WAV files there are, or how important this functionality is. But there are already user report of libsndfile not being able to open mp3 files, because libsndfile found the WAV header is trying to open it that way.

I'm not sure how accurate the codec-delay or other fields of the MPEGLAYER3 header for files in the wild. But if libsndfile sets them correctly, it would be ideal if a file encoded by libsndfile hass the same sample count and offset when decoded by libsndfile.

sobukus:
@arthurt With codec delay you mean the encoder delay and padding? As mpg123 knows its own decoder delay.

`MPG123_START_FRAME` and `MPG123_DECODE_FRAMES` have a different purpose. They're like `dd skip=n count=m`.

Currently, libmpg123 sees WAV headers and skips them as junk explicitly. It did not occur to me that they contain valuable information. For actual files, libmpg123 gets the accurent size in frames from scanning the file (if you trigger `mpg123_scan()` or just decode through it).

I am not sure how much value this business has. When you nowadays encode a file (using lame in sndfile), it will have that info in the Info frame and it will be used by the decoder. I see the value only in possibly being able to gaplessly decode existing files that happen to have proper delay/padding info recorded.

Can you show me one example of such a pre-existing file? If they only contain something like `length = 1152*frame_count`, then there is no added value to just decoding without that knowledge …

sobukus:
Sorry, you mentioned the pre-encoded stream sample count. So the question is if the encoder delay has been correctly recorded.

sobukus:
To be clear: I am open to adding support either for MPEGLAYER2WAVEFORMAT or passing along the delay/padding info via extra API. But I want to be convinced that this is a thing first, not just something academic. The next release of mpg123 that turns mpg123_init() into a no-op is in the pipeline, so something should happen quickly along those lines if any support should appear with 1.27.0.

arthurt:
The FACT chunk is pretty straightforward. I have seen WAV+MPEG files with this set. Of course, the total sample count on it's own doesn't help if both the beginning or end padding are unknown. Maybe one was supposed to infer from the common encoders of the day?!

The codec delay seems pretty straightforward:

> Encoder delay in samples. If you do not know this value, set this structure member to zero. [1]

But, I have not yet seen a WAV+MPEG file with it non-zero. Further, I do not know if software that can playback WAV with MPEG (not including software that can playback MPEG, skipping the WAV header) correctly heed this parameter.

I think the 'canonical' implementation of this would be the ACM stuff for WinMM, circa mid 90s Windows.

---

The first goals is that a WAV+MPEG file written by libsndfile has the same sample count and is not offset when read back by libsndfile.

The second lessor goal is that a WAV+MPEG file as read back by libsndfile has the same length and offset as other software.


To accomplish the first goal I think there are two ways:

1) Always write a Xing/Info header in the MPEG data chunk of the WAV file libsndfile creates, rely on that for total length and padding. (Still write correct values in the wav header, but only for posterity or some other imagined other software.)

2) Don't write an Xing/Info header, write and parse the total sample count and encoder delay from the WAV headers.

The first method is the easiest, and also works for mpg123 reading the output file without having to parse WAV headers itself. But, while the WAV+MPEG spec seems to exist after VBR encoders were common, I'm not sure that putting a Xing/Info header frame in is expected, so it seems a bit non-standard (for this not-very-common-or-used-standard.)

The second method could be made to work, but it limits the number of software that I currently know of that could playback the files correctly to one: libsndfile. If libmpg123 adds WAV header parsing, it would be a redundancy between the two projects.

Compounding this, we don't know currently if other software heeds the codec delay, and if it does, how it does it.

Let me do some testing on what existing WAV playback software that can handle MPEG do and get back to you.

---

[1] https://docs.microsoft.com/en-us/windows/win32/api/mmreg/ns-mmreg-mpeglayer3waveformat

evpobr:
@arthurt ?

makitsune:
oh..? i'm currious where this went

evpobr:
Hi @makitsune . This feauture has low priority now. @arthurt says he will implement it later, maybe in 1.1.1.

PR #724, opened 2021-03-31
Programs: Add MPEG functionality, Vorbis, Opus fixes

arthurt:
A few small fixes for the programs.

 - (common) Add missing MPEG format handling in to string conversions.
 - (common) Fix a style error that was somehow committed already!?
 - (convert) Do floating point sample copying when either the source or destination is MPEG
 - (convert) Infer Vorbis subformat from `.ogg` suffix. This aligns with what Xiph.org now recommends. `.oga` still means "Ogg container with audio" and is subformat agnostic.
 - (common) Don't warn about Broadcast Chunk for MPEG WAV files. By the comment above, they are allowed by the RFC.
 - (info) Calculate dB properly for Vorbis and Opus.

One non-program fix that's been bugging me for a while:
 - For Opus, remove tabs injected into a pre-formated info-dump block of printfs, messing up the alignment.

### DIFF
--- a/programs/common.c
+++ b/programs/common.c
@@ -73,7 +73,7 @@ sfe_copy_data_fp (SNDFILE *outfile, SNDFILE *infile, int channels, int normalize
 			{	data [k] /= max ;
 
 				if (!isfinite (data [k])) /* infinite or NaN */
-					return 1;
+					return 1 ;
 				}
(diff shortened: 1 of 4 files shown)


evpobr:
@arthurt , it is yours, merge it when you're ready.

PR #727, opened 2021-04-04
[cmake] Separate API and ABI versions

evpobr:


CMakeLists.txt
```diff
@@ -17,6 +17,15 @@ endif ()
...
 project(libsndfile VERSION 1.0.31)
 
+#
+# ABI version of library.
+#
+
+set (SNDFILE_ABI_VERSION_MAJOR 1)
+set (SNDFILE_ABI_VERSION_MINOR 0)
+set (SNDFILE_ABI_VERSION_PATCH 31)
+set (SNDFILE_ABI_VERSION "${SNDFILE_ABI_VERSION_MAJOR}.${SNDFILE_ABI_VERSION_MINOR}.${SNDFILE_ABI_VERSION_PATCH}")
```
SoapGentoo, line 27: how do you want to bump this? whats the procedure?

evpobr: This exact version of so binary (after so). Formula (if I'm not mistaken):

    c-a:r:a

See also: http://blog.asleson.org/2014/07/08/libtool-library-versioning-version-info-currentrevisionage/

SoapGentoo: so the idea is to imitate whatever comes out of libtool?

evpobr:
@SoapGentoo , this is first bug preventing us from using prereleases.

The second is in Autotools build toolchain. Makefile generates symbols files with `create_symbols.py` using `$(VERSION)` variable, which is API version.

I think it is the source of that bug:

> /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld:../../src/Symbols.gnu-binutils:3: ignoring invalid character `-' in script

CMakeLists.txt
evpobr: CMake just uses different encoding schema, but generates the same ABI version as libtool.

Actually, CMake's approach looks similar to SemVer.

SoapGentoo: CMake's `VERSION` has no effect on the SONAME, just on the filenames and macOS' `-compatibility-version` arguments (which actually follow SemVer). If you want to make the install trees identical, then I'm fine with it, I just don't think emulating libtool's arcane model is a great idea otherwise. But we can go with this.

evpobr: Isn't it?:

```
[113/143] : && /usr/bin/gcc -fPIC -Wall -Wextra -O3 -DNDEBUG  -Wl,--version-script,/home/runner/work/libsndfile/libsndfile/build/src/Symbols.gnu-binutils -shared -Wl,-soname,libsndfile.so.1 -o libsndfile.so.1.0.31
```

SoapGentoo:
```sh
$ tree staging/usr/local/lib64/
staging/usr/local/lib64/
├── cmake
│   └── SndFile
│       ├── SndFileConfig.cmake
│       ├── SndFileConfigVersion.cmake
│       ├── SndFileTargets.cmake
│       └── SndFileTargets-noconfig.cmake
├── libsndfile.so -> libsndfile.so.1
├── libsndfile.so.1 -> libsndfile.so.1.0.31
├── libsndfile.so.1.0.31
└── pkgconfig
    └── sndfile.pc

3 directories, 8 files
$ readelf -d staging/usr/local/lib64/libsndfile.so | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libsndfile.so.1]
```
clearly the `VERSION` is never part of the SONAME.

evpobr: Interesting. Can you provide output for Autotools?

SoapGentoo:
```sh
$ tree staging-autotools/usr/local/lib/
staging-autotools/usr/local/lib/
├── libsndfile.a
├── libsndfile.la
├── libsndfile.so -> libsndfile.so.1.0.31
├── libsndfile.so.1 -> libsndfile.so.1.0.31
├── libsndfile.so.1.0.31
└── pkgconfig
    └── sndfile.pc

1 directory, 6 files
$ readelf -d staging-autotools/usr/local/lib/libsndfile.so | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libsndfile.so.1]
```
just remember the `.1.0.31` suffix just **happens** to be the same as the version because I went out of my way to make it so. Most projects don't have matching suffixes (just look at ffmpeg).

SoapGentoo (approved):

Issue #731, opened 2021-04-13
Heap-buffer-overflow in src/flac.c:274:41 in flac_buffer_copy


yuawn:
Hi, I found a vulnerability in current master [34bd39b](https://github.com/libsndfile/libsndfile/commit/34bd39b9b513ce6267d7b1b7cd2af56697eb2f24).

There is a heap-buffer-overflow in src/flac.c:274:41 in flac_buffer_copy.

The vulnerability can lead to heap-based buffer overflow via a crafted sound file, and potentially control heap data by forge `buffer` content to perform heap exploitation.

To reproduce on x86-64 Ubuntu 20.04.2 with clang-11:
```bash
export CC=clang-11
export CFLAGS='-g -fsanitize=address'

cd libsndfile
./autogen.sh
mkdir build
cd build
../configure --disable-shared
make

./programs/sndfile-convert ./heap_overflow_poc ./out.wav
```

PoC: [heap_overflow_poc.gz](https://github.com/libsndfile/libsndfile/files/6305061/heap_overflow_poc.gz)

ASAN report:
```
=================================================================
==3712127==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x621000001310 at pc 0x00000054e793 bp 0x7ffe0d1368b0 sp 0x7ffe0d1368a8
READ of size 4 at 0x621000001310 thread T0
    #0 0x54e792 in flac_buffer_copy /home/yuawn/fuzz-targets/libsndfile/master/libsndfile/build/../src/flac.c:274:41
    #1 0x54fc83 in flac_read_loop /home/yuawn/fuzz-targets/libsndfile/master/libsndfile/build/../src/flac.c:946:3
    #2 0x549be9 in flac_read_flac2i /home/yuawn/fuzz-targets/libsndfile/master/libsndfile/build/../src/flac.c:997:13
    #3 0x512fb4 in sf_readf_int /home/yuawn/fuzz-targets/libsndfile/master/libsndfile/build/../src/sndfile.c:1946:10
    #4 0x4ffae5 in sfe_copy_data_int /home/yuawn/fuzz-targets/libsndfile/master/libsndfile/build/../programs/common.c:94:22
    #5 0x4fee78 in main /home/yuawn/fuzz-targets/libsndfile/master/libsndfile/build/../programs/sndfile-convert.c:367:3
    #6 0x7f1f78add0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
[...]

PR #732, opened 2021-04-14
flac: Fix improper buffer reusing

yuawn:
See #731 .

### DIFF
--- a/src/flac.c
+++ b/src/flac.c
@@ -62,7 +62,6 @@ typedef struct
 	FLAC__StreamMetadata *metadata ;
 
 	const int32_t * const * wbuffer ;
-	unsigned wbuffer_size ;
 	int32_t * rbuffer [FLAC__MAX_CHANNELS] ;
 
 	int32_t* encbuffer ;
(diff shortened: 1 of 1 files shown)


evpobr:
@arthurt , can you look at this?

evpobr:
It looks like @arthurt is busy, no activity in his repos.

@bobsayshilol, could you review it?

bobsayshilol (changes requested):
I can confirm that this PR makes the crash go away, however I don't believe that this is the correct solution to the problem.

The underlying issue is that the return value of `FLAC__stream_decoder_process_single()` isn't correctly handled here: https://github.com/libsndfile/libsndfile/blob/4c30646abf7834e406f7e2429c70bc254e18beab/src/flac.c#L957-L967

According to the documentation at https://xiph.org/flac/api/group__flac__stream__decoder.html#ga9d6df4a39892c05955122cf7f987f856 it returns `false if any fatal read, write, or memory allocation error occurred (meaning decoding must stop), else true` so we should be dropping into the branch below and setting `pflac->frame = NULL` so that the next read doesn't read from a now bogus frame, which is currently what's happening since `pflac->frame` is pointing to an outdated/corrupt frame inside `FLAC`.

evpobr:
@yuawn , would you be so kind as to update the pull request?

yuawn:
Hi @evpobr,

Thank @bobsayshilol, I have updated the pull request.
This commit can fix the root cause of this security bug, and it should be merge and release as soon as possible, because the vulnerability affects wild range of release version.

evpobr:
Funny, the fix is so simple. @bobsayshilol , approve?

bobsayshilol (approved):
LGTM. Note that `psf_log_printf()` isn't being called like in the branch below though, but functionally it's now correct.

yuawn:
@evpobr, done!

evpobr:
Thank you all for your help!

PR #735, opened 2021-04-20
Update sndfile.pc for MPEG support

evpobr:


configure.ac
```diff
@@ -423,6 +421,7 @@ AS_IF([test -n "$PKG_CONFIG"], [
 				AS_IF([test "x$ac_cv_lame$ac_cv_mpg123" = "xyesyes"], [
 						enable_mpeg="yes"
 						HAVE_MPEG=1
+						EXTERNAL_MPEG_REQUIRE="libmpg123"
```
SoapGentoo, line 424: @evpobr had to fix the variable, and used `EXTERNAL_MPEG_REQUIRE` to be more consistent with the rest.

CMakeLists.txt
```diff
@@ -219,6 +219,23 @@ if (INSTALL_PKGCONFIG_MODULE)
...
+		if (CMAKE_IMPORT_LIBRARY_PREFIX)
+			string (REGEX REPLACE "^${CMAKE_IMPORT_LIBRARY_PREFIX}" "" LAME_WE_NO_PREFIX ${LAME_WE})
+			if (LAME_HIP_LIBRARY)
+				string (REGEX REPLACE "^${CMAKE_IMPORT_LIBRARY_PREFIX}" "" LAME_HIP_WE_NO_PREFIX ${LAME_HIP_WE})
+			endif ()
+		endif ()
+		set (MPEG_LIBS "-l${LAME_WE_NO_PREFIX}")
+		if (LAME_HIP_LIBRARY)
+			set (MPEG_LIBS "${MPEG_LIBS} -l${LAME_HIP_WE}")
+		endif ()
```
SoapGentoo, line 237: I'm unhappy with `MPEG_LIBS`: If you drop `libmpg123` as a `Requires.private` into the `.pc`, that obviates the need for specifying `MPEG_LIBS` in `Libs.private`. I strongly suggest you get rid of the `Libs.private` line.

SoapGentoo (changes requested):

CMakeLists.txt
evpobr: No, it is not libmpg123, it is LAME. It has no `.pc` file.

SoapGentoo: oh sorry, missed that, :+1:

evpobr:
@SoapGentoo ?

SoapGentoo (approved):

PR #736, opened 2021-04-25
Fix heap-buffer-overflow in copyPredictorTo24

bobsayshilol:
This PR fixes the crash seen in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27503 by re-enabling code that shouldn't have been disabled in the original import. AFAICT this is the original code: https://github.com/macosforge/alac/blob/master/codec/ALACDecoder.cpp#L588-L593. However since we're importing this we would want it to be in release mode with this check enabled by default. Without this check `channelIndex` can increase without bounds and that leads to writing past the end of the output buffer.


evpobr:
@bobsayshilol , thanks!

PR #737, opened 2021-05-06
float32, double64: Simplify read/write code

evpobr:
It looks so ridiculously simple that I can't believe everything is correct here.

It is very interesting to know the opinions.

Not tested on big endian processors.

### DIFF
--- a/src/double64.c
+++ b/src/double64.c
@@ -282,148 +282,69 @@ double64_init	(SF_PRIVATE *psf)
 
 double
 double64_be_read (const unsigned char *cptr)
-{	int		exponent, negative, upper, lower ;
+{
 	double	dvalue ;
+	uint8_t *dbytes = (uint8_t *) &dvalue ;
 
-	negative = (cptr [0] & 0x80) ? 1 : 0 ;
-	exponent = ((cptr [0] & 0x7F) << 4) | ((cptr [1] >> 4) & 0xF) ;
-
-	/* Might not have a 64 bit long, so load the mantissa into a double. */
-	upper = (((cptr [1] & 0xF) << 24) | (cptr [2] << 16) | (cptr [3] << 8) | cptr [4]) ;
-	lower = (cptr [5] << 16) | (cptr [6] << 8) | cptr [7] ;
-
-	if (exponent == 0 && upper == 0 && lower == 0)
-		return 0.0 ;
-
-	dvalue = upper + lower / ((double) 0x1000000) ;
-	dvalue += 0x10000000 ;
-
-	exponent = exponent - 0x3FF ;
-
-	dvalue = dvalue / ((double) 0x10000000) ;
-
-	if (negative)
-		dvalue *= -1 ;
-
-	if (exponent > 0)
-		dvalue *= pow (2.0, exponent) ;
-	else if (exponent < 0)
-		dvalue /= pow (2.0, abs (exponent)) ;
+	dbytes [0] = cptr [7] ;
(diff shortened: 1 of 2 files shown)


src/double64.c
```diff
@@ -282,148 +282,69 @@ double64_init	(SF_PRIVATE *psf)
 
 double
 double64_be_read (const unsigned char *cptr)
-{	int		exponent, negative, upper, lower ;
+{
 	double	dvalue ;
+	uint8_t *dbytes = (uint8_t *) &dvalue ;
```
SoapGentoo, line 287: don't use `uint8_t`, which is not guaranteed to alias other pointers. Always use `unsigned char*`, which can always be used to alias other types.

And yes, this is the [Rob Pike pattern](https://commandcenter.blogspot.com/2012/04/byte-order-fallacy.html), and on modern GCC and Clang, this often optimizes straight down to `MOVBE` on Haswell type platforms.

src/double64.c
```diff
@@ -282,148 +282,69 @@ double64_init	(SF_PRIVATE *psf)
...
-	else if (exponent < 0)
-		dvalue /= pow (2.0, abs (exponent)) ;
+	dbytes [0] = cptr [7] ;
+	dbytes [1] = cptr [6] ;
+	dbytes [2] = cptr [5] ;
+	dbytes [3] = cptr [4] ;
+	dbytes [4] = cptr [3] ;
+	dbytes [5] = cptr [2] ;
+	dbytes [6] = cptr [1] ;
+	dbytes [7] = cptr [0] ;
```
SoapGentoo, line 296: given that you aren't doing endianness targetting here, why not use `memcpy` instead? GCC/Clang optimize `memcpy` away on arguments with compile-time sizes.

src/double64.c
```diff
@@ -282,148 +282,69 @@ double64_init	(SF_PRIVATE *psf)
...
-	else if (exponent < 0)
-		dvalue /= pow (2.0, abs (exponent)) ;
+	dbytes [0] = cptr [0] ;
+	dbytes [1] = cptr [1] ;
+	dbytes [2] = cptr [2] ;
+	dbytes [3] = cptr [3] ;
+	dbytes [4] = cptr [4] ;
+	dbytes [5] = cptr [5] ;
+	dbytes [6] = cptr [6] ;
+	dbytes [7] = cptr [7] ;
```
SoapGentoo, line 314: memcpy

src/double64.c
```diff
@@ -282,148 +282,69 @@ double64_init	(SF_PRIVATE *psf)
...
+	uint8_t *dbytes = (uint8_t *) &in ;
+
+	out[0] = dbytes[7];
+	out[1] = dbytes[6];
+	out[2] = dbytes[5];
+	out[3] = dbytes[4];
+	out[4] = dbytes[3];
+	out[5] = dbytes[2];
+	out[6] = dbytes[1];
+	out[7] = dbytes[0];
```
SoapGentoo, line 332: memcpy

SoapGentoo (changes requested):
Do memcpy for the rest too. Once you're finished, I can test this on our SPARC (BE) machines

evpobr:
Now I already doubt whether it needs to be optimized. We use it in `psf_binheader_readf` and `psf_binheader_writef`, that's ok. But this code is also used in `float32.c` and `double64.c` with `FLOAT_BROKEN_BE` and `DOUBLE_BROKEN_BE`. What is that, do you know?

PR #743, opened 2021-05-16
Fix CMake shared library ABI version on Apple platforms

evpobr:
Now ABI version must be compatible with Autotools builds. Note that this change requires CMake >= 3.17 for building dylib on Apple platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([issue 27366](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27366)).
 * Integer overflow in `nms_adpcm_update`(), credit to OSS-Fuzz
   ([issue 25522](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25522)).
+* ABI version incompatibility between Autotools and CMake build on Apple
+  platforms.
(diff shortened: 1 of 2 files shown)


evpobr:
@SoapGentoo , do you have the opportunity to test this on macOS?

SoapGentoo:
master:
`@rpath/libsndfile.1.dylib (compatibility version 1.0.0, current version 1.0.31)`
fix-cmake-apple-platform-dylib-abi:
`@rpath/libsndfile.1.dylib (compatibility version 2.0.0, current version 2.31.0)`

SoapGentoo (approved):

evpobr:
@SoapGentoo , any ideas about libexpat/libexpat#55?

SoapGentoo:
I still believe fundamentally that trying to make libtool and CMake bug-compatible is a lost cause. We should realyl work towards getting rid of Autotools.

hartwork:
@SoapGentoo for Expat, it should be noted that getting rid of Autotools means a circular dependency loop CMake -> Expat -> CMake. At least for Expat, Autotools is not likely to go away.

SoapGentoo:
I get that, and expat is a bit special in this regard, something libsndfile doesn't have.

hartwork:
@SoapGentoo nevermind, I was not intending to speak for cases other than Expat :+1:

PR #745, opened 2021-05-19
autotools: Use ABI version for symbol scripts

evpobr:


configure.ac
```diff
@@ -84,7 +84,20 @@ dnl   5. If any interfaces have been added since the last public release, then i
...
+m4_define([lt_revision], [0])
+m4_define([lt_age], [1])
+
+dnl This is ABI version for linker scripts, CMake uses the same format for
+dnl VERSION property of shared library.
+dnl The formula is: c:r:a -> c-a:a:r
+
+m4_define([abi_version_major], [m4_eval(lt_current - lt_age)])
+m4_define([abi_version_minor], [lt_age])
+m4_define([abi_version_patch], [lt_revision])
```
SoapGentoo, line 100: you're assuming ABI versioning follows API versioning?

evpobr: No. We still pass `SHARED_VERSION_INFO` in libtool compatible format, e.g. `1.31.0`.

`ABI_VERSION` is required to write linux symbols script properly. If we pass `SHARED_VERSION_INFO` there, symbol script `Symbols.gnu-binutils` will contain `libsndfile.so.1.31` section and this is wrong. Actually it will leed to ABI break.

So we need another variable with libtool version reencoded to compatible format. It is the same as version part after `.so` extension, e.g `1.31.0` -> `1.0.31`. And this will work as expected with our `create_symbols_file.py` script.

For next `1.1.0beta1` release:

* API version - 1.1.0beta1
* libtool version - 1:32:0

SoapGentoo: I'll test it tomorrow

evpobr: @SoapGentoo , ok?

SoapGentoo:
```
--- master
+++ use-abi-version-for-symbol-scripts
@@ -1,4 +1,4 @@
 └── usr
     └── local
         ├── bin
@@ -18,9 +18,9 @@
         ├── lib
         │   ├── libsndfile.a
         │   ├── libsndfile.la
-        │   ├── libsndfile.so -> libsndfile.so.1.0.31
-        │   ├── libsndfile.so.1 -> libsndfile.so.1.0.31
-        │   ├── libsndfile.so.1.0.31
+        │   ├── libsndfile.so -> libsndfile.so.1.1.0
+        │   ├── libsndfile.so.1 -> libsndfile.so.1.1.0
+        │   ├── libsndfile.so.1.1.0
         │   └── pkgconfig
         │       └── sndfile.pc
         └── share
```
That's intentional? because you haven't bumped the actual `PACKAGE_VERSION` yet?

evpobr: Oh, test values, i will fix that. Thanks you.

evpobr: Now it should be ok.

SoapGentoo: yup, checked

SoapGentoo (approved):

SoapGentoo:
@evpobr can we delete your feature branch?

evpobr:
Done.

PR #748, opened 2021-05-25
Read CMake ABI version from configure.ac

evpobr:
This change will make release process easier - bump `lt_current`, `lt_revision` and `lt_age` in `configure.ac` and CMake will update ABI version on configure or build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,23 +18,6 @@ endif ()
 project(libsndfile VERSION 1.0.31)
 
 #
-# ABI version of library.
-#
-
-set (SNDFILE_ABI_VERSION_MAJOR 1)
-set (SNDFILE_ABI_VERSION_MINOR 0)
-set (SNDFILE_ABI_VERSION_PATCH 31)
-set (SNDFILE_ABI_VERSION "${SNDFILE_ABI_VERSION_MAJOR}.${SNDFILE_ABI_VERSION_MINOR}.${SNDFILE_ABI_VERSION_PATCH}")
-
-#
-# Apple platform current and compatibility versions.
-#
-
-math (EXPR SNDFILE_MACHO_CURRENT_VERSION_MAJOR "${SNDFILE_ABI_VERSION_MAJOR} + ${SNDFILE_ABI_VERSION_MINOR} + 1")
-set (SNDFILE_MACHO_CURRENT_VERSION "${SNDFILE_MACHO_CURRENT_VERSION_MAJOR}.${SNDFILE_ABI_VERSION_PATCH}.0")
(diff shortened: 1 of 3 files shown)


SoapGentoo:
I'm not a super fan of creating such monstrosities in CMake, but w/e

SoapGentoo (approved):

evpobr:
> I'm not a super fan of creating such monstrosities in CMake, but w/e

In fact, every second Find module does something similar:

This is from standard `FindZLIB.cmake`:

```cmake
if(ZLIB_INCLUDE_DIR AND EXISTS "${ZLIB_INCLUDE_DIR}/zlib.h")
    file(STRINGS "${ZLIB_INCLUDE_DIR}/zlib.h" ZLIB_H REGEX "^#define ZLIB_VERSION \"[^\"]*\"$")

    string(REGEX REPLACE "^.*ZLIB_VERSION \"([0-9]+).*$" "\\1" ZLIB_VERSION_MAJOR "${ZLIB_H}")
    string(REGEX REPLACE "^.*ZLIB_VERSION \"[0-9]+\\.([0-9]+).*$" "\\1" ZLIB_VERSION_MINOR  "${ZLIB_H}")
    string(REGEX REPLACE "^.*ZLIB_VERSION \"[0-9]+\\.[0-9]+\\.([0-9]+).*$" "\\1" ZLIB_VERSION_PATCH "${ZLIB_H}")
    set(ZLIB_VERSION_STRING "${ZLIB_VERSION_MAJOR}.${ZLIB_VERSION_MINOR}.${ZLIB_VERSION_PATCH}")

    # only append a TWEAK version if it exists:
    set(ZLIB_VERSION_TWEAK "")
    if( "${ZLIB_H}" MATCHES "ZLIB_VERSION \"[0-9]+\\.[0-9]+\\.[0-9]+\\.([0-9]+)")
        set(ZLIB_VERSION_TWEAK "${CMAKE_MATCH_1}")
        string(APPEND ZLIB_VERSION_STRING ".${ZLIB_VERSION_TWEAK}")
    endif()

    set(ZLIB_MAJOR_VERSION "${ZLIB_VERSION_MAJOR}")
    set(ZLIB_MINOR_VERSION "${ZLIB_VERSION_MINOR}")
    set(ZLIB_PATCH_VERSION "${ZLIB_VERSION_PATCH}")
endif()
```

SoapGentoo:
> > I'm not a super fan of creating such monstrosities in CMake, but w/e
> 
> In fact, every second Find module does something similar:
> 
> This is from standard `FindZLIB.cmake`:

Yes, and this is a reason I hate all of these `Find*.cmake` modules, because they're extremely brittle, they break constantly, and they get all the edge cases wrong, something that literally never happens with `pkg-config`. If I had a dollar for every time some CMake upgrade broke some module in Gentoo, I could retire by now.

PR #760, opened 2021-06-18
Use forward for loops

evpobr:


evpobr:
@arthurt , i've got error using forward for loop in mpeg_decode.c: f2d_array() with `-O3`:

https://godbolt.org/z/c1j14sva6

```
139: Test command: D:\source\repos\libsndfile\build\mpeg_test.exe
139: Test timeout computed to be: 1500
139:     mpeg_short_test                : mpeg_short.mp3 .......... ok
139:     mpeg_seek_test                 : mpeg_short.mp3 .......... ok
139:     mpeg_int_test                  : mpeg_int.mp3 ............ ok
139:     mpeg_seek_test                 : mpeg_int.mp3 ............ ok
139:     mpeg_float_test                : mpeg_float.mp3 .......... ok
139:     mpeg_seek_test                 : mpeg_float.mp3 .......... ok
139:     mpeg_double_test               : mpeg_double.mp3 ......... ok
139:     mpeg_seek_test                 : mpeg_double.mp3 .........
139:
139: Line 252 : Error at index 0, got  0, should be  4.98663e-05.
139:
1/1 Test #139: mpeg_test ........................***Failed    0.09 sec
```

How could it be? I see the code is auto vectorized (that's what i want from this PR), maybe the result is slightly different? 

Compiler is GCC 10.3, MSVC 2017 also fails.

SoapGentoo:
I really like this, but this requires some close reviewing, on the weekend?

evpobr:
> I really like this, but this requires some close reviewing, on the weekend?

No problem.

arthurt:
Pretty straightforward. Only caution I have, and I don't know if it's true, but if a caller of a function like s2alaw where the destination type is larger than the source is converting a buffer in place by passing the same pointer for both a source and destination, it only works if the loop iterates backwards.

Consider libc's memmove(), essentially it chooses what direction it iterates based on the sort order of the passed buffers' addresses.

I'm not saying that's we need to insert logic like that into these loops, just a caution to check callers of the functions with these loops to identify if any copy from a buffer back to itself. If so we need to fix those cases, either leaving the loop direction backwards or using an intermediate buffer.

Not sure what's going on with the mpeg loop. I've been really busy this week getting ready for a trip, and now I'm leaving away from cellular service for a week, but if it can wait, I'll have a look when I'm back :)

evpobr:
> Not sure what's going on with the mpeg loop. I've been really busy this week getting ready for a trip, and now I'm leaving away from cellular service for a week, but if it can wait, I'll have a look when I'm back :)

Ok, waiting.

evpobr:
@SoapGentoo , @arthurt do you have time to review it? :smile:

SoapGentoo:
@evpobr ...I can't reproduce the failure? can you give me your exact `CFLAGS`? Your test output looks like CTest.
`CFLAGS="-O3 -march=native" CXXFLAGS="-O3 -march=native" ./configure && make check` (on GCC 10):

```sh
    mpeg_short_test                : mpeg_short.mp3 .......... ok
    mpeg_seek_test                 : mpeg_short.mp3 .......... ok
    mpeg_int_test                  : mpeg_int.mp3 ............ ok
    mpeg_seek_test                 : mpeg_int.mp3 ............ ok
    mpeg_float_test                : mpeg_float.mp3 .......... ok
    mpeg_seek_test                 : mpeg_float.mp3 .......... ok
    mpeg_double_test               : mpeg_double.mp3 ......... ok
    mpeg_seek_test                 : mpeg_double.mp3 ......... ok
    mpeg_stereo_seek_test          : mpeg_seek.mp3 ........... ok
```

`CFLAGS="-O3 -march=native" CXXFLAGS="-O3 -march=native" cmake ..`:
```sh
        Start 139: mpeg_test
139/143 Test #139: mpeg_test ........................   Passed    0.02 sec
```

SoapGentoo (approved):

evpobr:
This pull request contains no buggy code, did you convert backward loops in mpeg files?

SoapGentoo:
@evpobr I see where the issue is:
`f2d_array (const float *src, int count, double *dest, double normfact);`
so the first argument has to be `float*`
line 209:
`f2d_array ((float *) ptr, done, ptr, normfact) ;`
but `ptr` is a `double*`, i.e., this pointer cast is undefined behaviour. That you are seeing this failure mode is due to the fact that forward loops are easier to optimize for compilers than the crappy backwards loop from before. This needs to be fixed properly.

evpobr:
> @evpobr I see where the issue is:
> `f2d_array (const float *src, int count, double *dest, double normfact);`
> so the first argument has to be `float*`
> line 209:
> `f2d_array ((float *) ptr, done, ptr, normfact) ;`
> but `ptr` is a `double*`, i.e., this pointer cast is undefined behaviour. That you are seeing this failure mode is due to the fact that forward loops are easier to optimize for compilers than the crappy backwards loop from before. This needs to be fixed properly.

I've never seen this trick before, @arthurt surprises me again. :smile: This loop should be backward, or all the code should be rewritten completely. Not sure it is bug or it is feature, but it works.  I'll leave it to @arthurt.

SoapGentoo:
@evpobr could you prepare a branch `on top` of this one with the changes for mpeg too? I'll look at that failure then

evpobr:
> @evpobr ...I can't reproduce the failure? can you give me your exact `CFLAGS`? Your test output looks like CTest.
> `CFLAGS="-O3 -march=native" CXXFLAGS="-O3 -march=native" ./configure && make check` (on GCC 10):
> 
> ```shell
>     mpeg_short_test                : mpeg_short.mp3 .......... ok
>     mpeg_seek_test                 : mpeg_short.mp3 .......... ok
>     mpeg_int_test                  : mpeg_int.mp3 ............ ok
>     mpeg_seek_test                 : mpeg_int.mp3 ............ ok
>     mpeg_float_test                : mpeg_float.mp3 .......... ok
>     mpeg_seek_test                 : mpeg_float.mp3 .......... ok
>     mpeg_double_test               : mpeg_double.mp3 ......... ok
>     mpeg_seek_test                 : mpeg_double.mp3 ......... ok
>     mpeg_stereo_seek_test          : mpeg_seek.mp3 ........... ok
> ```
> 
> `CFLAGS="-O3 -march=native" CXXFLAGS="-O3 -march=native" cmake ..`:
> 
> ```shell
>         Start 139: mpeg_test
> 139/143 Test #139: mpeg_test ........................   Passed    0.02 sec
> ```

It was simple CMake's Release build profile (uses `-O3` optimization), nothing special.

evpobr:
> @evpobr could you prepare a branch **on top** of this one with the changes for mpeg too? I'll look at that failure then

https://github.com/libsndfile/libsndfile/tree/use-forward-for-loops-with-mpeg

SoapGentoo:
yup, besides the undefined behavior due to the strict-aliasing violations, the fundamental problem is that of the in-place expansion:
```
|-- f0 --|-- f1 --|-- f2 --|-- f3 --|-- f4 --|-- f5 --|
                           v
                       f2d_array
                           v
|------ d0 -------|------ d1 -------|------ d2 -------|
```
The only way to do that safely is from the back, doing it from the front would start killing input data before you loaded it for processing.

I think this function should be made to use different buffers and not do this inplace.

evpobr:
In any case, i would like to hear what @arthurt  has to say.

arthurt:
> Only caution I have, and I don't know if it's true, but if a caller of a function like s2alaw where the destination type is larger than the source is converting a buffer in place by passing the same pointer for both a source and destination, it only works if the loop iterates backwards.

I guess I should have know that that was true... I wrote such a routine then forgot :laughing:

There are two ways to fix this like you say: Retain the backwards looping (and also fix the strict aliasing rule breaking using a pointer to a union of float and double), or use another intermediate buffer (which is how `mpeg_dec_read_s()` and `mpeg_dec_read_i()` do it already.)

I elected for the latter, and also cleaned the loops up a bit. [Consider pulling this branch](https://github.com/arthurt/libsndfile/tree/use-forward-for-loops-with-mpeg) into your branch.

arthurt:
Note to self: If this gets merged first, #747 will need fixing.

arthurt:
Also,

> Not sure what's going on with the mpeg loop.

because at the time I was looking this branch, which did not modify the mpeg code :-P

arthurt (approved):

evpobr:
Thanks everyone!

PR #775, opened 2021-09-08
Fix duplicated case value of ESTRPIPE on BSD

papadave66:
It was built with an error in FreeBSD:
```
programs/sndfile-play.c: In function 'alsa_write_float':
programs/sndfile-play.c:331:4: error: duplicate case value programs/sndfile-play.c:293:4: note: previously used here
```
because this file has defined `ESTRPIPE` as `EPIPE` in: `/usr/local/include/alsa/pcm.h:#define ESTRPIPE EPIPE`

### DIFF
--- a/programs/sndfile-play.c
+++ b/programs/sndfile-play.c
@@ -328,10 +328,12 @@ alsa_write_float (snd_pcm_t *alsa_dev, float *data, int frames, int channels)
 					return 0 ;
 					break ;
 
+#if defined ESTRPIPE && ESTRPIPE != EPIPE
 			case -ESTRPIPE :
 					fprintf (stderr, "alsa_write_float: Suspend event.n") ;
 					return 0 ;
 					break ;
+#endif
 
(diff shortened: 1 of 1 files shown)


evpobr (approved):

SoapGentoo (approved):

SoapGentoo:
would be nice having CI on FreeBSD too, but I doubt that is doable without self-hosted runners on Github

evpobr:
@papadave66 , thanks for pull request.

papadave66:
> @papadave66 , thanks for pull request.

You are welcome.

evpobr:
> would be nice having CI on FreeBSD too, but I doubt that is doable without self-hosted runners on Github

I don't mind, but I have no idea how to implement it.

diizzyy:
fwiw, Gerbera project uses sr.ht which may be a viable solution
See https://github.com/gerbera/gerbera/pull/1918 for example

PR #778, opened 2021-09-21
configure.ac: Remove stack smash protection

SoapGentoo:
* This is something deployers should enable if they want it.


evpobr:
Yep, MinGW doesn't support it correctly. So this option is useful, thank you.

evpobr (approved):

PR #787, opened 2021-11-03
Fix integer overflow in psf_log_printf()

evpobr:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28441

Credit to OSS-Fuzz.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -103,8 +103,8 @@ log_putchar (SF_PRIVATE *psf, char ch)
 void
 psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 {	va_list		ap ;
-	uint32_t	u ;
-	int			d, tens, shift, width, width_specifier, left_align, slen, precision ;
+	uint32_t	u, tens ;
+	int			d, shift, width, width_specifier, left_align, slen, precision ;
 	char		c, *strptr, istr [5], lead_char, sign_char ;
 
 	va_start (ap, format) ;
@@ -186,15 +186,19 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 					d = va_arg (ap, int) ;
 
 					if (d < 0)
-					{	d = -d ;
-						sign_char = '-' ;
+					{	sign_char = '-' ;
 						if (lead_char != '0' && left_align == SF_FALSE)
 							width_specifier -- ;
-						} ;
+
+						u = - ((unsigned) d) ;
+						}
+					else
+					{	u = (unsigned) d ;
+						}
 
 					tens = 1 ;
 					width = 1 ;
-					while (d / tens >= 10)
+					while (u / tens >= 10)
 					{	tens *= 10 ;
 						width ++ ;
 						} ;
@@ -224,8 +228,8 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 							log_putchar (psf, lead_char) ;
 
 					while (tens > 0)
-					{	log_putchar (psf, '0' + d / tens) ;
-						d %= tens ;
+					{	log_putchar (psf, '0' + u / tens) ;
+						u %= tens ;
 						tens /= 10 ;
 						} ;
 
@@ -234,7 +238,8 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 					break ;
 
 			case 'D': /* sf_count_t */
-					{	sf_count_t		D, Tens ;
+					{	sf_count_t	D ;
+						uint64_t	U, Tens ;
 
 						D = va_arg (ap, sf_count_t) ;
 
@@ -244,13 +249,19 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 							log_putchar (psf, '0') ;
 							break ;
 							}
-						if (D < 0)
-						{	log_putchar (psf, '-') ;
-							D = -D ;
-							} ;
+						else
+						{	if (D < 0)
+							{	log_putchar (psf, '-') ;
+								U = -((uint64_t) D) ;
+								}
+							else
+							{	U = (uint64_t) D;
+								}
+							}
+
 						Tens = 1 ;
 						width = 1 ;
-						while (D / Tens >= 10)
+						while (U / Tens >= 10)
 						{	Tens *= 10 ;
 							width ++ ;
 							} ;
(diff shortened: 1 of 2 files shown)


evpobr:
Windows build fail is not related to code change:

```
Error: Failed to download from mirror set:
https://gitlab.xiph.org/xiph/ogg/-/archive/v1.3.5/ogg-v1.3.5.tar.gz: WinHttpSendRequest() failed: 12002
```

src/common.c
```diff
@@ -186,15 +186,19 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
...
 					if (d < 0)
-					{	d = -d ;
-						sign_char = '-' ;
+					{	sign_char = '-' ;
 						if (lead_char != '0' && left_align == SF_FALSE)
 							width_specifier -- ;
-						} ;
+						
+						u = (d != INT32_MIN) ? -d : INT32_MAX ;
+						}
```
SoapGentoo, line 194: another solution would be to cast to unsigned, perform the subtraction, and cast back to signed. While that invokes implementation-defined behaviour, I'm not aware of any platform/compiler that doesn't just take the bit pattern unchanged.

SoapGentoo (approved):

src/common.c
evpobr: I hadn't thought of that. I hadn't thought of that. I'll leave it as it is for now. Let's see if @arthurt has any thoughts on this.

arthurt: Humm, it's also wrong. abs(INT32_MIN) != abs(INT32_MAX) (on two's-compliment hardware, aka, everything.)

Signed-negative-to-unsigned is well defined, as is unsigned unary minus, so doing `-((unsigned)d)` as @SoapGentoo  suggests should be okay.

src/common.c
```diff
@@ -186,15 +186,19 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 					d = va_arg (ap, int) ;
 
 					if (d < 0)
-					{	d = -d ;
-						sign_char = '-' ;
+					{	sign_char = '-' ;
 						if (lead_char != '0' && left_align == SF_FALSE)
 							width_specifier -- ;
-						} ;
+
```
arthurt, line 192: (Nit-pick) (Not showing up in GitLab, even with "Hide Whitespace Changes" disabled.) Trailing whitespace added on line 192.

src/common.c
evpobr: > Humm, it's also wrong. abs(INT32_MIN) != abs(INT32_MAX) (on two's-compliment hardware, aka, everything.)

I don't understand guys, where is UB in line 193? Maybe you didn't notice that `u` is **unsigned**.

arthurt: > Maybe you didn't notice that u is unsigned.

Yeah, but I'm not confident in the order of operations from a plain `u = -d`. Is that equivalent to `d = -d; u = d` or `u = d; u = -u`? We want to avoid the former, use the latter as unsigned overflow is defined.

Force the issue by using `u = -((unsigned)d)`.

arthurt:
Humm, I'm not confident there isn't a bigger issue here. The `d` format specifier is for values of type `int`. Remember that _technically_ all that's required is that `sizeof(int) >= sizeof(short)`. It just so happens that `sizeof(int) == 4` is pretty much ubiquitous.

I see in commit 3e91aaf7a Erik changed `u` from `unsigned int` to `uint32_t`, as part of work to "fix problems found by [the] clang sanitizer." This however requires the assumption that `sizeof(int) == 4` for this code to work correctly. I'm not sure using uint32_t is correct here. Maybe the clang sanitizer was throwing up false positives? Is there any platform/reason why `sizeof(int) != sizeof(unsigned int)` ?

src/common.c
evpobr: Okay. Then why not `(unsigned) -d`?  We change sign of signed integer to positive, the value is safe because it is `!= INT32_MIN` branch of ternary operator. Then we cast positive signed integer to unsigned. Looks sane,

arthurt: We, given the predicate of the outer ternary conditional, `(unsigned) -d` is safe, but:

Firstly, it's an unnecessary conditional. The arithmetic required is possible and defined without a conditional. (But this isn't really that important, performance of this function is not of concern.)

Mostly because the second half of the ternary is wrong. `-(INT32_MIN) != INT32_MAX`. (Assuming two's compliment) INT32_MIN = `-2147483648`, INT32_MAX = `214783647`. So, using that conditional, the value -2147483648 would be printed as -2147483647 by this function. (2147483648 is representable as an unsigned int, though, all the way up to 4294967295, aka -1)

You could fix that by doing
```
u = (d != INT32_MIN) ? (unsigned) -d : ((unsigned)INT32_MAX) + 1;
```
or more correctly with
```
u = (d != INT32_MIN) ? (unsigned) -d : -((unsigned)INT32_MIN);
```
at which point it is completely equivalent to:
```
u = -((unsigned)d);
```

evpobr:
> Humm, I'm not confident there isn't a bigger issue here. The `d` format specifier is for values of type `int`. Remember that _technically_ all that's required is that `sizeof(int) >= sizeof(short)`. It just so happens that `sizeof(int) == 4` is pretty much ubiquitous.
> 
> I see in commit [3e91aaf](https://github.com/libsndfile/libsndfile/commit/3e91aaf7aa6c9e961edbc5d1f3a529ac1637c2b2) Erik changed `u` from `unsigned int` to `uint32_t`, as part of work to "fix problems found by [the] clang sanitizer." This however requires the assumption that `sizeof(int) == 4` for this code to work correctly. I'm not sure using uint32_t is correct here. Maybe the clang sanitizer was throwing up false positives? Is there any platform/reason why `sizeof(int) != sizeof(unsigned int)` ?

I have no idea. We have a lot of code that uses basic C types, where we need to use fixed-size types. Even in the public interface. If we are going to make version 2.0, we will need to fix this.

evpobr:
I understand the review is over. Time to merge.

src/common.c
evpobr: @arthurt, @SoapGentoo  , I have problems with this code after upgrading to Visual Studio 2022.

here

https://github.com/libsndfile/libsndfile/blob/33e765ccba9a0eb225694fdbf9e299683a8338ee/src/common.c#L193

and here:

https://github.com/libsndfile/libsndfile/blob/33e765ccba9a0eb225694fdbf9e299683a8338ee/src/common.c#L255

With VS2022 on [UWP platform](https://docs.microsoft.com/en-us/windows/uwp/get-started/) [SDL checks](https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks?view=msvc-170) are enabled by default and [C4146 warning](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=msvc-170) turned into an error.

Fix this

```c
u = - ((unsigned) d) ; 
```

on

```c
(uint32_t) u = - ((unsigned) d) ; 
```

is correct?

Issue #662, opened 2020-12-05
Wishlist: Support Ogg Opus Ambisonics


kcat:
Libopus 1.3 added support for ambisonics, using channel mapping families 2 and 3 to differentiate it from normal multichannel files. This is described in [RFC 8486](https://tools.ietf.org/html/rfc8486). Libsndfile already has some support for ambisonics, via `amb` files which are (uncompressed) `wav` files with some special headers (a WAVE_FORMAT_EXTENSIBLE format header with unique subformat GUIDs and a blank channel mask). An application can detect such an ambisonic wav/amb file by querying
```c
if(sf_command(handle, SFC_WAVEX_GET_AMBISONIC, NULL, 0) == SF_AMBISONIC_B_FORMAT)
```

However, this can't be directly used for Opus ambisonic files because Opus employs a different, more widely-used mapping which isn't compatible. The channels are in a different order and they use different weighting -- specifically, `amb` files use Furse-Malham (FuMa) channel order and weights, while Opus adopts ACN ordering and SN3D weighting ("AmbiX"). Playing FuMa style content as if it was AmbiX, or AmbiX style content as if it was FuMa, will sound wrong, so an application needs to be aware of what it's getting to interpret it properly. Or alternatively, libsndfile could automatically convert to/from the style it already uses, although that would need more processing to fix up, add more quantization noise, and stick the API with the outdated FuMa style.

PR #797, opened 2021-11-25
Add support for Opus ambisonics and other channel mapping families

aib:
This is my first time contributing to this project so please feel free to leave a review.

This basically adds Opus ambisonics support, granted you have a B-format encoder. Tested using libspatialaudio and YouTube's player.

Would satisfy #662.


evpobr:
Hi @aib . Looks good to me. But let's ask @arthurt .

arthurt:
Hi @aib, thanks for contributing!

The code itself looks good, and thanks for writing tests for the new features.

However, I think this is the first half of support. libsndfile already supports one other ambisonic format, so we should try and add generic commands for setting/getting ambisonic configuration that aren't tied to the subformat.

I wonder, is it possible that SFC_GET_CHANNEL_MAP_INFO could be extended to relay the required information? I've not worked with ambisonic formats, so I don't know much about the mapping  #662 notes that the amb format uses a different map and weights than Opus. Is adding new SF_CHANNEL_MAP_AMBISONIC_* required or sufficient?

If SFC_GET/SET_CHANNEL_MAP_INFO cannot be usefully extended, maybe we could add another subformat-independent command. SFC_GET/SET_AMBISONIC_INFO with a new data type?

More generally, our Opus support doesn't currently have SFC_SET_CHANNEL_MAP_INFO and SFC_GET_CHANNEL_MAP_INFO, which we really should do.

aib:
First: I don't mind my efforts going to waste as it is the risk I took by submitting a PR out of the blue. Thank you for taking the time to review and respond.

Looking through the documentation, `SFC_WAVEX_GET_AMBISONIC` seems to set a flag in the file header which allows a many-channel wavex file to be recognised as carrying B-format data.

Opus' channel mapping families 2 and 3 do the same.

I don't see why #662 mentions channel ordering and normalization as they seem to be out of scope for the library. If there was a format that required channel 0 to be right and 1 left—or let's say an entirely different speaker configuration with channel 0 as "forward right-ish" and 1 as "back-ish left", would that mean any changes to the library? I'm genuinely asking because I'm not very familiar with it. From how I've used it so far, if there was a format that carried R/L channels in reverse order, I would just make sure that the first channel is played and recorded on the right and the second channel on the left. There is no channel order processing in the library, is there?

I haven't worked with wavex, but from what I can tell: If a file has 4 channels and subformat `KSDATAFORMAT_SUBTYPE_IEEE_FLOAT`, it is presumed to have data for 4 speakers (which 4 specified in `dwChannelMask`, in a fixed order). If it has subformat `SUBTYPE_AMBISONIC_B_FORMAT_IEEE_FLOAT`, it is presumed to have B-format data for 4 channels: W, X, Y, Z (in that order).

This is the same in Opus: If a file has 4 channels and channel mapping family 1, they are presumed to be FL, FR, RL, RR in that order (https://datatracker.ietf.org/doc/html/rfc7845#section-5.1.1.2). If it has 4 channels and channel mapping family 2, they are presumed to be BW, BY, BZ, BX with SN3D normalization (https://datatracker.ietf.org/doc/html/rfc8486#section-3.1).

What's a possible source of confusion is that "channel mapping" (as opposed to channel mapping *family* which I've been talking about) seems to be a different and almost orthogonal concept in Opus: Streams in the file are input channels and there is an expected number of output channels. The channel mapping array allows an output channel to be fed from an input channel and duplicates are allowed; e.g. `[0, 1, 0, 1]` would have the same sound playing on the 1st and 3rd channels and another one on the 2nd and 4th.

The library seems to simply place here an array `[0, 1, ..., n-1]`, effectively disabling this feature.

As for `CHANNEL_MAP_INFO`, how is it used in formats with fixed channel ordering? I'll check the code whenever I have time next but I assume it is read only and returns a fixed structure on GET? This much could be possible in Opus, too, as long as we're talking output channels (or the [0..n-1] channel mapping which makes them equivalent to the input channels).

If you ask my opinion, I don't see much point in trying to entwine the completely different ambisonics formats. In fact, between formats with differing normalizations, even the existing `SF_CHANNEL_MAP_AMBISONIC_B_{W,X,Y,Z}` aren't interchangeable.

aib:
> Despite the added commands not including OPUS in the names, the idea of a channel mapping family is exclusive to Opus.

This was bugging me, too. Fixed with an amend.

arthurt:
> I don't see why #662 mentions channel ordering and normalization as they seem to be out of scope for the library. ... There is no channel order processing in the library, is there?

That's correct. libsndfile doesn't include any remappers or resamplers, and should strive just to abstract the file-format details of the audio files, and not process the audio itself. This gets blurred a bit by lossy formats, particularly Opus. 

For terminology, libsndfile considers channel-mapping to be a mapping between a read/write channel and an identifier. So channel 0 is "left", channel 1 is "right" for stereo for example. Just identifiers.

As you say, Opus also has stream-mappings, mapping streams of multistream-Opus with channels. A multistream-Opus encoder is free to choose how to represent `n` channels by `n` or fewer streams, and the streams themselves can be single or joined, containing one or two channels.

A Channel Mapping Family from the Opus documentation is not just how to map identifiers to channels, but also how to encode channels-to-stream mappings.

> The library seems to simply place here an array [0, 1, ..., n-1], effectively disabling this feature.

libsndfile doesn't decide the stream mapping and coupling, and instead gets libopus to decide. See https://github.com/xiph/opus/blob/a8e6a77c5fe0c37aa6788f939f24f8cd22ae2652/src/opus_multistream_encoder.c#L528 . We just send libopus the number of input channels and a channel mapping family, and it decides the rest.

Currently, libsndfile does the following:

- If the number of channels is one or two, use channel mapping family 0. No stream mapping table in the header, one implicit single or coupled stream.

- If the number of channels is three to eight, use channel mapping family 1. Opus multistream encoder assumes Vorbis channel ordering, and couples the streams to be efficient for that: (https://github.com/xiph/opus/blob/a8e6a77c5fe0c37aa6788f939f24f8cd22ae2652/src/opus_multistream_encoder.c#L53)

- If the number of channels is more than eight, use channel mapping family 255: Opus multistream encoder just chooses n single streams

This can be improved, and ambisonic support should be added. Ambisonics aside, this current code has the following shortcomings:

- An implicit channel-mapping is being used, influencing compression-quality, but is not being reported through `SFC_GET_CHANNEL_MAP_INFO`. 

- There are actually at least two possible channel mappings for each channel count <= 8: family 0/1 (vorbis mapping) and family 255 (no mapping). But there is no way for a user to choose this. Add in Ambisonics and there are more choices for channel mappings.

I'd prefer to solve the problem in the most subformat generic way. We need some way for callers to express their intent, and then we infer the Opus channel mapping family from that. Maybe it could be by setting a specific channel map and we infer, or maybe it is a call like 'SFC_SET_AMBISONIC_X"? Hummm....

One design principle to keep in mind is if you write a file with one configuration, reading it again should return the same configuration. So only metadata and configuration of a format that is written unambiguously into the file should be settable and gettable by users.

---

I just noticed that your changes as-is won't work. The channel mapping family is chosen and the stream-mappings decided in `ogg_opus_setup_encoder()`, before `sf_open()` has returned and you have a `SNDFILE *` to call `sf_command()` on. The changes will change the channel mapping family written in the header, as the header write comes later, but the encoder will not change it's behaviour and the stream mappings are fixed.

aib:
*Headdesk* I missed that `opus_multistream_surround_encoder_init` had a mapping family parameter. That changes everything, of course.

> libsndfile doesn't decide the stream mapping and coupling, and instead gets libopus to decide. See https://github.com/xiph/opus/blob/a8e6a77c5fe0c37aa6788f939f24f8cd22ae2652/src/opus_multistream_encoder.c#L528 . We just send libopus the number of input channels and a channel mapping family, and it decides the rest.

This is very annoying, especially given that the "surround" versions have pretty much the same parameters, use `mapping` for output instead of input, and are missing [documentation](https://opus-codec.org/docs/opus_api-1.3.1/group__opus__multistream.html#ga8d87c08ecbe6ed4ce1ede4e58718b621).

> I'd prefer to solve the problem in the most subformat generic way. We need some way for callers to express their intent, and then we infer the Opus channel mapping family from that. Maybe it could be by setting a specific channel map and we infer, or maybe it is a call like 'SFC_SET_AMBISONIC_X"? Hummm....

For what it's worth, I agree. I tried some things, but was ultimately shot down by my unfamiliarity with the library's design or just the unfamiliar sequence of steps. In any case, though, it looks like encoder creation will have to be repeated or delayed, as `ogg_opus_setup_encoder` is called early in the process, no?

In any case, I don't see much point in going through with this PR. I learned a lot about Opus, especially with your comments filling in the gaps, thank you. If you can think of a way we could salvage our efforts here or I can be of help to the library, let me know. Especially if there's a decision to use something like `SFC_SET_AMBISONIC_X`, I would love to take another crack at it.

arthurt:
> This is very annoying, especially given that the "surround" versions have pretty much the same parameters, use mapping for output instead of input, and are missing documentation.

Yeah, not being documented is annoying. I inferred their usage from reading the `libopusfile` documentation.

Interestingly, `opus_multistream_surround_encoder_init()` also has code in it for handling channel mapping family 2. Again, it looks like you only need to specify a channel count and it builds a set of streams:
- https://github.com/xiph/opus/blob/a8e6a77c5fe0c37aa6788f939f24f8cd22ae2652/src/opus_multistream_encoder.c#L104
- https://github.com/xiph/opus/blob/a8e6a77c5fe0c37aa6788f939f24f8cd22ae2652/src/opus_multistream_encoder.c#L559

> In any case, though, it looks like encoder creation will have to be repeated or delayed, as ogg_opus_setup_encoder is called early in the process, no?

Yup. It is called during the user's call to `sf_open()` so it can fail-fast. However, it would be possible, (and probably necessary for changing the channel mapping family), to delay the encoder creation or recreate it on demand. Other codecs in libsndfile do this (MP3 as an example) and return any error from failing to create the encoder on the first call to `sf_write*()`.

> In any case, I don't see much point in going through with this PR. I learned a lot about Opus, especially with your comments filling in the gaps, thank you. If you can think of a way we could salvage our efforts here or I can be of help to the library, let me know. Especially if there's a decision to use something like SFC_SET_AMBISONIC_X, I would love to take another crack at it.

Thanks for your contribution and discussion! Ambisonics wasn't on my radar, and we welcome contributions. Happy hacking!

Issue #808, opened 2022-02-05
building for android with AndroidStudio and CMake


scar20:
Hi,
It would be nice to have a built for android. I'm on Windows with only AndroidStudio for coding just to say that I'm not fluent with build tools. Fortunately, I could find a sample app https://github.com/sheraz-nadeem/oboe_recorder_sample that use libsndfile and I've been able to duplicate it, build it and add my own method to get a short buffer from libsndfile. It build libsndfile 1.0.29 as a static lib included in a shared lib referred by the app.

Now, I tried to rebuild with the latest version of libsndfile to benefit all the hard work (and thanks a lot for this) to integrate mp3 support. This time it fail. From the entry point CMakeList.txt, I adapted include_directory to reflect the new structure of the distribution, from:
```
### ADD LIBSNDFILE LIBRARY
set(LIB_SND_FILE_DIR ../../../../libsndfile)
add_subdirectory(${LIB_SND_FILE_DIR} ./sndfile)
include_directories(${LIB_SND_FILE_DIR}/src)
```
to:
```
### ADD LIBSNDFILE LIBRARY
set(LIB_SND_FILE_DIR ../../../../libsndfile)
add_subdirectory(${LIB_SND_FILE_DIR} ./sndfile)
include_directories(${LIB_SND_FILE_DIR}/include)
```
It fail at:
`ninja: error: 'C:/Users/Sylvain/AndroidStudioProjects/libsndfileProjects/SndFileLib2/libsndfile/cmake/sqlite/CMakeAutoGenScript.cmake', needed by 'sndfile/tests/utils.c', missing and no known rule to make it`
I tried naively to just copy CMakeAutoGenScript.cmake into the sqlite folder but that didn't work either.

I also tried to build a shared lib from libsndfile (the app is destined will need to have libsndfile as a shared lib) using another sample app as the template: https://github.com/android/ndk-samples/tree/master/hello-libs which comprise a "lib generator". Modifying the code I successful build the libsndfile 1.1.0 (I hava a libsndfile.so) but the lib is not working as expected: it seem the generation of sndfile.h got screwed up (all the sf_count are "unknown"). I have to say that I bluntly added SHARED as the second argument of the libsndfile static build block:
```
add_library (sndfile
		SHARED
	src/sfconfig.h
        etc...
```

and added this block at the end of the "static"(now shared) build which replicate the others install in the sample app:
```
## copy out the lib binary and remove generated files
[...]

PR #810, opened 2022-02-21
cmake: Fix build on some systems

evpobr:
Related to #808.


SoapGentoo (approved):

evpobr:
Still doesn't work properly. I guess it is better to remove that CMake module.

PR #812, opened 2022-03-01
configure.ac: trigger error & stop if not all Xiph libs available

commodo:
The context here is that I sent a patch to Yocto, to update the library
from 1.0.28 to 1.0.31 and move to the new URLs for libsndfile.

However, in the meantime, libsndfile also supports libopus and does not
allow support for libvorbis, libflac and libogg to be included if either of
these libs (and libopus) does not exist.

That's fine, but it allows for weird/silent failures that can slip through
when you're a package maintainer.

Which is what happened here:
  https://lists.openembedded.org/g/openembedded-core/message/162558?p=%2C%2C%2C20%2C0%2C0%2C0%3A%3Acreated%2C0%2Clibsndfile1%2C20%2C2%2C0%2C89367260

So, the idea is to propose that this build configuration should trigger a
fail that can be seen at build time and not be caught off-guard.

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -392,8 +392,7 @@ AS_IF([test -n "$PKG_CONFIG"], [
 					AC_MSG_WARN([[***]])
 					AC_MSG_WARN([[*** Unfortunately, for ease of maintenance, the external libs]])
 					AC_MSG_WARN([[*** are an all or nothing affair.]])
-					AS_ECHO([""])
-					enable_external_libs=no
+					AC_MSG_ERROR(["You requested external libraries, but not all were found"])
 				])
 	])
 


SoapGentoo:
@evpobr what do you think? I'm fine with this, it just makes it slightly harder for people who just download libsndfile tarballs and run the configure scripts in those.

evpobr:
Hi @commodo.

evpobr:
Hi @SoapGentoo . Let's try. Will you take this PR?

evpobr (approved):

configure.ac
```diff
@@ -392,8 +392,7 @@ AS_IF([test -n "$PKG_CONFIG"], [
 					AC_MSG_WARN([[***]])
 					AC_MSG_WARN([[*** Unfortunately, for ease of maintenance, the external libs]])
 					AC_MSG_WARN([[*** are an all or nothing affair.]])
-					AS_ECHO([""])
-					enable_external_libs=no
+					AC_MSG_ERROR(["Unsupported configuration. Will stop build, to prevent silent packaging failures"])
```
SoapGentoo, line 395:
```suggestion
					AC_MSG_ERROR(["You requested external libraries, but not all were found"])
```

SoapGentoo (changes requested):

configure.ac
commodo: Sure
Will do that tomorrow.

It’s time to put the kids to sleep in my TZ and then myself.

Thanks
Alex

commodo:
ping on this;

so, after talking a bit with OE, there is a preference to have a more granular control over these options;

i can also do it instead of this;

but i would like a consensus on this

arthurt:
> I was thinking about splitting the configuration of the Xiph libraries, ENABLE_VORBIS, ENABLE_FLAC and so on. Is it better to solve the problem this way?

I was in favour of this option, but @erikd was not. See #400 .

evpobr:
I misunderstood the essence of this PR. Now I see that it turns a valid configuration without external Xiph libraries into an error. I think this is wrong.

In fact, both Autotools and CMake report at the end of configuration that support for external libraries is disabled. I always thought that was enough. So it can hardly be said that this is a problem on our part.

However, the idea of separate options is a good one. Therefore, I suggest closing this request pool and scheduling separate options for configuring Xiph libraries. We will ask @erikd if he is still against this decision. @erikd?

Now we'll wait a bit for what he says. If we do not answer, we will assume that he does not mind.

commodo:
> I misunderstood the essence of this PR. Now I see that it turns a valid configuration without external Xiph libraries into an error. I think this is wrong.
> 

so, the configuration is valid (i.e. not external Xiph libs)

but downstream this can lead to silent failures ;
i.e. any one of these libs can be missing (or misconfigured which can be similar to missing) and then all get disabled;

> In fact, both Autotools and CMake report at the end of configuration that support for external libraries is disabled. I always thought that was enough. So it can hardly be said that this is a problem on our part.

i understand the reasoning for this approach (i.e. maintenance/testing effort)
but it throws this (testing/maintenance) cost downstream;
now, i am not saying that libsndfile must pick up this cost; everyone is free to choose their own costs;

a scenario to this all-or-nothing mechanism, is that another Xiph would get added to libsndfile, then it is upgraded downstream, then the old Xiph libs go missing silently;

> However, the idea of separate options is a good one. Therefore, I suggest closing this request pool and scheduling separate options for configuring Xiph libraries. We will ask @erikd if he is still against this decision. @erikd?

i also like the separate options idea;
i assumed that as a no-go option, as the current all-or-nothing mechanism implies this 

i was not aware of #400 ;
but that looks like it mostly has implemented the solution;
so, OpenEmbedded is on board with this solution;
the discussion can be seen here:   https://lists.openembedded.org/g/openembedded-core/topic/89367260#162558

in the OE case, the preference is to not move Speex into OE-core (it's packaged somewhere else)

> Now we'll wait a bit for what he says. If we do not answer, we will assume that he does not mind.

i think 7 days timeout (for a reply) sounds reasonable (i would say)
so, if i may come back at around that time;

commodo:
so, Github says I'm late 4 days to give a ping back :)
it's been 11 days since my last comment, mentioning that I'd come back in 7 days 

any thoughts on how to move ahead with this?

evpobr:
So we will separate options.

commodo:
Ack.

Will close this.

Anything needed from my side?

I can try to pick-up the patch from @arthurt and do an updated version.

evpobr:
No, thanks for report.

commodo:
> No, thanks for report.

thank you :)
i'm mostly glad i managed to turn a fail in a win (for my first patch with open-embedded :p )

kanavin:
@commodo Can you then resurrect #400, rebase and reopen it then? Otherwise, we'll have to continue carring the patch submitted here, which is not a win.

arthurt:
I don't mind re-doing the #ifdef work of #400 again, I know the code surface pretty well, but the autotools and CMake stuff isn't my favorite, and it's changed a bit since then.

commodo:
(I'm a little slow to do things these days)

I can also do it based on @arthurt 's work.
I don't mind the autotools + CMake stuff;

PR #819, opened 2022-03-31
Support skipping large wav chunks on stdin.

swesterfeld:
Some wav files can have rather large chunks with chunk types that are
not supported by libsndfile at all. For those, header_seek will just
skip the whole chunk.

This commit makes skipping the chunk work even if the input is stdin
(which doesn't support psf_fseek), by reading over the header using
psf_fread in that case.

______________________

I have a wav file here which has these chunks:
```
$ rifftree -s bug1.wav
RIFF(WAVE)-> (39882456 Bytes)
            fmt ; (16 Bytes)
            sav1; (42 Bytes)
            sav2; (4 Bytes)
            bext; (642 Bytes)
            pad ; (130308 Bytes)
            data; (39751396 Bytes)
```
Libsndfile can read this file correctly if I provide a filename to open it. If I read it from stdin, it doesn't work. I have added some debugging (`SFC_GET_LOG_INFO`) to my code, and here is the result for stdin:
```
$ cat bug1.wav | audiowmark add - out.wav f0
[[File : -
Length : unknown
RIFF : 39882460
WAVE
fmt  : 16
  Format        : 0x1 => WAVE_FORMAT_PCM
  Channels      : 2
  Sample Rate   : 44100
  Block Align   : 4
  Bit Width     : 16
  Bytes/sec     : 176400
*** sav1 : 42 (unknown marker)
*** sav2 : 4 (unknown marker)
bext : 642
*** pad  : 130308 (unknown marker)
Request for header allocation of 260616 denied.
psf_fseek : pipe seek to value other than pipeoffset
Have 0 marker at position 764 (0x2FC).
]]
audiowmark: error opening -: Error in WAV file. No 'data' chunk marker.
```
I tracked this down, the problem is that libsndfile will try to skip over the huge unknown "pad " chunk using psf_fseek. But if the input is a pipe, seeking is not supported. So this commit fixes the problem by reading over the chunk data using psf_fread if the input is a pipe.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -925,8 +925,22 @@ header_seek (SF_PRIVATE *psf, sf_count_t position, int whence)
 
 			if (psf->header.indx + position > psf->header.len)
 			{	/* Need to jump this without caching it. */
+				position -= (psf->header.end - psf->header.indx) ;
 				psf->header.indx = psf->header.end ;
-				psf_fseek (psf, position, SEEK_CUR) ;
+				if (psf->is_pipe)
+				{
+					/* seeking is not supported on pipe input, so we read instead */
+					size_t skip = position ;
+					while (skip)
+					{	char junk [16 * 1024] ;
+						size_t to_skip = SF_MIN (skip, sizeof (junk)) ;
+						psf_fread (junk, 1, to_skip, psf) ;
+						skip -= to_skip ;
+						}
+					}
+				else
+				{	psf_fseek (psf, position, SEEK_CUR) ;
+					}
 				break ;
 				} ;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * CMake build from Autotools tarball (issue #816).
 * Build on UWP platform (issue #824).
 * Fix signed integer overflow (issue #785).
+* Skipping large wav chunks on stdin (PR #819).
 
 ## [1.1.0] - 2022-03-27
 


evpobr:
hi @swesterfeld .

evpobr (changes requested):
It looks good for me. Fix coding style and i will be happy.

arthurt (approved):

src/common.c
```diff
@@ -926,7 +926,21 @@ header_seek (SF_PRIVATE *psf, sf_count_t position, int whence)
 			if (psf->header.indx + position > psf->header.len)
 			{	/* Need to jump this without caching it. */
 				psf->header.indx = psf->header.end ;
-				psf_fseek (psf, position, SEEK_CUR) ;
+				if (psf->is_pipe)
+				{
+					/* seeking is not supported on pipe input, so we read instead */
+					size_t skip = position ;
```
arthurt, line 932: Small issue here (actually an existing bug.) A backwards header seek may have occurred in the buffer, meaning that there are bytes already read an sitting in `header.ptr` between `header.indx` and `header.end`. These bytes need to count against the total in `position` as they have already been read.

So before `psf->header.indx = psf->header.end`, it is needed that `position -= (header.end - header.indx)`.

I found this out when testing if this could help reading ID3v2 headed files from a pipe. Initally we read 12 bytes to try and identify the file header in `guess_file_type()`, so `header.end = 12`. Then `id3_skip()` seeks back to 0 and reads 10 bytes, so `header.indx = 10`. If the ID3 header excluding the parsed 10 bytes is another 10000 bytes long, the correct amount to skip is 10000 - 2 bytes, not 10000.

arthurt (changes requested):
Can I leverage your MR to fix the existing bug in header_seek. Add 
```
position -= (psf->header.end - psf->header.indx) ;
```
after line 927, just before
```
psf->header.indx = psf->header.end ;
``

File: src/common.c [hunk 2]
```diff
@@ -925,8 +925,22 @@ header_seek (SF_PRIVATE *psf, sf_count_t position, int whence)
 
 			if (psf->header.indx + position > psf->header.len)
 			{	/* Need to jump this without caching it. */
+				position -= (psf->header.end - psf->header.indx) ;
```
evpobr, line 928: @arthurt , just added line of code.

arthurt (approved):

PR #820, opened 2022-03-31
Handle EAGAIN when reading files

nigoroll:
Apparently this can happen even when the file descriptor is in blocking mode, as seen in this strace:

```
176595 read(0,  <unfinished ...>
176595 <... read resumed>0x7ffdf39ace94, 2092) = -1 EAGAIN (Resource temporarily unavailable)
```

The case at hand happens with [elbestream](https://code.uplex.de/misc/elbestream), which has been written as intermediate relief for sound processing programs which fail stream processing on some inputs.

When stdin of the program using libsndfile is handed over to [vmsplice()](https://code.uplex.de/misc/elbestream/blob/e42437c85432ef042743e3004e7c52295e3ae35f/elbestream.c#L168) the `EAGAIN` is seen repeatedly, though not always.

### DIFF
--- a/src/file_io.c
+++ b/src/file_io.c
@@ -361,7 +361,7 @@ psf_fread (void *ptr, sf_count_t bytes, sf_count_t items, SF_PRIVATE *psf)
 		count = read (psf->file.filedes, ((char*) ptr) + total, (size_t) count) ;
 
 		if (count == -1)
-		{	if (errno == EINTR)
+		{	if (errno == EINTR || errno == EAGAIN)
 				continue ;
 
 			psf_log_syserr (psf, errno) ;


nigoroll:
Ironically, the issue which I tried to work around is #819

evpobr:
Hi @nigoroll .

Explain please, does #819 help with your problem?

nigoroll:
Hi @evpobr ,
I ran into #819 which caused processing to fail with that particular input when read from a pipe, but not when read from a file.
So I tried to use [elbestream](https://code.uplex.de/misc/elbestream) to selectively provide the piped input buffered into a file when processing fails, but that also failed due to the `EAGAIN`. So my original problem is solved with #819, but that does not in any way change the fact that Linux can return `EAGAIN` when reading from a pipe.
So this PR should be justified independent of the the particular conditions under which the issue was seen first. I only mentioned the link because I found it ironic, but technically the mention would not have been necessary and I apologize if that has caused unnecessary confusion.

nigoroll:
@evpobr removed additional cstyle commit

evpobr (approved):

arthurt:
In general, libsndfile assumes all IO is blocking.

Looping on `EAGAIN` without any kind of wait is not a good idea. If this code were merged and a user opened a non-blocking pipe, 100% CPU spin loops could happen inside of psf_read()

If the FD you are using from vmsplice which is in blocking-mode is returning EAGAIN on read, that sounds like a bug with glibc or the kernel. Unfortunately, while looping on EAGAIN seems to fix the issue in this case, adding a loop on EAGAIN is a bad idea in the more likely scenario of a users opening a non-blocking pipe.

As a work-around, you could use `sf_open_virtual` and the `sf_vio` wrappers. That allows you to implement seek, read, write, tell as required.

This would even help with #819, as you could implement `seek(<positive offset>, SEEK_CUR, ... )` as a discarding-read.

```
struct my_vio_userdata {
    ...
};

my_sf_vio_read(void *ptr, sf_count_t count, void *userdata) {
    struct my_vio_userdata *m = (struct my_vio_userdata *) userdata;
     ...
}

nigoroll:
@arthurt I agree with your point about the loop scenario should user code put a pipe into non-blocking mode. However the argument seems inconsistent with

> In general, libsndfile assumes all IO is blocking.

That said, I will dig deeper to understand the exact origin of the `EAGAIN`.

PR #840, opened 2022-05-29
Fix build errors on UWP platform with Visual Studio 2022

evpobr:
I ran into problems when updating the [Vcpkg](https://github.com/microsoft/vcpkg) port to 1.1.0, specifically the build for the [UWP platform](https://docs.microsoft.com/en-us/windows/uwp/get-started/universal-application-platform-guide) was completed with [warning C4146: "unary minus operator applied to unsigned type, result still unsigned"](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?f1url=%3FappId%3DDev16IDEF1%26l%3DRU-RU%26k%3Dk(C4146)%26rd%3Dtrue&view=msvc-170).

Although UWP is also partly Windows, but a lot of things are done differently there and from some point the warning was translated into an error.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -190,7 +190,7 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 						if (lead_char != '0' && left_align == SF_FALSE)
 							width_specifier -- ;
 
-						u = - ((unsigned) d) ;
+						u = - (int32_t) ((unsigned) d) ;
 						}
 					else
 					{	u = (unsigned) d ;
@@ -252,7 +252,7 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 						else
 						{	if (D < 0)
 							{	log_putchar (psf, '-') ;
-								U = -((uint64_t) D) ;
+								U = - (int64_t) ((uint64_t) D) ;
 								}
 							else
 							{	U = (uint64_t) D;


arthurt:
Umm, that sounds like a bug with the compiler really.

> [warning C4146: "unary minus operator applied to unsigned type, result still unsigned"](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?f1url=%3FappId%3DDev16IDEF1%26l%3DRU-RU%26k%3Dk(C4146)%26rd%3Dtrue&view=msvc-170).

> Unsigned types can hold only non-negative values, so unary minus (negation) usually doesn't make sense when applied to an unsigned type. Both the operand and the result are non-negative.

So, Microsoft is reinterpreting the C spec now?

Unary minus on unsigned types is

1) Defined
2) What we want in this context.

Remember the discussion in #787 . Unary minus on unsigned integers is defined and does what we want, and doing a cast through a sign results in an undefined overflow.

arthurt:
Yup

> The C4146 warning is just a bit of Microsoft nonsense. Look it up to understand it, then ignore it. Perfectly correct programs can emit this warning.

arthurt:
It may be possible to disable treating this warning as an error by disabling "[security development lifecycle](https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks)" additional warnings.

Looking at all the enabled warning-as-error behaviour it lists, they are all overly broad catch-all warnings that flag legitimate usage.

It might be possible to disable just C4146 with `/we4146-`. I don't have Visual Studio to test with.

evpobr:
> It may be possible to disable treating this warning as an error by disabling "[security development lifecycle](https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks)" additional warnings.
> 
> Looking at all the enabled warning-as-error behaviour it lists, they are all overly broad catch-all warnings that flag legitimate usage.
> 
> It might be possible to disable just C4146 with `/we4146-`. I don't have Visual Studio to test with.

Yes, i did it for vcpkg port.

Ok then, let's leave it as it is.

Issue #850, opened 2022-07-08
Version 1.1.0 WAV file maximum sample rate limited to 655350 sps.


drmpeg:
I'm a developer on the GNU Radio project and we use libsndfile for our WAV file handling. We've noticed that version 1.1.0 limits the maximum sample rate of WAV files to 655350 samples per second.

Previous versions did not have this limit (1.0.28 works to at least 1 Gsps).

This is a severe bug for us. Users routinely create WAV files at above 1 Msps (which is actually a low sample rate for Software Defined Radios).

drmpeg:
Here's the problematic line.

https://github.com/libsndfile/libsndfile/blob/master/src/sndfile.c#L2911

From commit https://github.com/libsndfile/libsndfile/commit/b1d642078f1278f8a1be20ff1f9206415fd34100

PR #851, opened 2022-07-09
Removed samplerate < SF_MAX_SAMPLERATE requirement

justacec:
This is a PR to remove the SF_MAX_SAMPLERATE test which was recently added to the `validate_sfinfo` function.  This corrects issue #850.  

This PR just simply removes the maximum sample rate check which might not be the desired approach to fix this issue.  However, as issue #850 indicates, the SDR community will typically have much higher bitrates and therefore there really is not upper limit on this.


evpobr (commented):
Hi @drmpeg, @justacec. Looks reasonable to me. I have one question, is there any documentation about the maximum practically used sample rate of WAV files?

drmpeg:
@evpobr The maximum practical sample rate is difficult to pin down. There are SDRs capable of 500 MS/s, although writing a file at that rate would be a challenge.

https://www.ettus.com/all-products/usrp-x410/

There are many hobbyist targeted SDRs that can sample to 61.44 MS/s.

https://www.ettus.com/all-products/usrp-b200mini/

But the only limitation for WAV files would be the 32-bit unsigned int storing sample rate, or 4,294,967,295 samples per second.

justacec:
I like the idea of the last point which is to base the limit on the technical capability of the format.  Libsndfile is focused on the format and therefore it seems reasonable that the scope for decisions should stop at that point.  Makes things easy.  :)

evpobr:
> But the only limitation for WAV files would be the 32-bit unsigned int storing sample rate, or 4,294,967,295 samples per second.

But `SF_INFO::samplerate` is signed int :smile:

evpobr (changes requested):
So checking samplerate like i did is pretty useless. Please remove also `SF_MAX_SAMPLERATE`. It would be nice to add note to changelog.

justacec:
Ahh.  Good points.  I will make those changes in just a bit.

evpobr (commented):
Add "### Removed" header above entry, change title to "Maximum samplerate checking"

evpobr:
Damn merge conflicts.

evpobr:
Let me fix it.

evpobr (approved):

evpobr:
I'll wait until tomorrow, and if there are no objections, I'll merge this PR. Thanks again!

Issue #431, opened 2018-12-10
Failure to read flac files of 3 channels longer than 5592405 samples


gokikamukow:
On line 976, 996, 1016, 1036 of flac.c a magic number 0x1000000 is used to determine a maximum read chunk size for a file. If the total bytes requested exceeds this value then 0x1000000 is used as the read size instead of the number of samples requested. The number of samples requested must be a multiple of the number of channels for the read operation to be successful. If it is not then an error "Error: pflac->remain 16777216    channels 3" is printed and the operation fails, returning no data. 0x1000000 is divisible by 1, 2, 4, 8 etc but will fail for non-power of 2 channel files that are sufficiently long to trigger this condition.

The fix should be fairly straightforward, simply round the magic number down to the next multiple of the number of channels.

For those finding this issue before a fix is made the workaround is to never read more than 0x1000000 / n_channels frames at a time.

dofuuz:
Fixed with #858

PR #858, opened 2022-07-26
Read more than 2^24 samples from FLAC file with non-power-of-two channels

bemoody:
As discussed in issue #431: when reading a FLAC file, the `sf_read_*` or `sf_readf_*` functions will fail if the total number of samples exceeds 0x1000000, and the number of channels is not a power of two.

This is because (e.g.) `flac_read_flac2s` will invoke `flac_read_loop` with a buffer size of at most 0x1000000, and `flac_read_loop` requires the size to be a multiple of the number of channels.

This logic was introduced with commit d81614ca9068959884a6db3b78b8892e76055956, which unfortunately does not have a very detailed commit message.  That commit does two things that appear unrelated:

(1) It ensures that `flac_buffer_copy` makes a private copy of the input data.

(2) It uses `unsigned` counters in `flac_read_loop` and elsewhere, in place of `sf_count_t`.

The first change is necessary because of how libFLAC works (you cannot continue accessing the data passed to `sf_flac_write_callback` after `sf_flac_write_callback` returns), and presumably this was the "valgrind error" in question.

The second change is not easy to explain, but one possible benefit is that `sf_count_t` might have been only 32 bits, and this logic permits reading more than 2^31 samples in that case.  (Which would only have made sense on truly crazy platforms like 64-bit Windows, but hey.)  Nowadays, as far as I can tell, `sf_count_t` is always at least 64 bits, so this should no longer be an issue.

At any rate: a simple fix for issue #431 is to change the internal size limit to a value that is divisible by every possible number of channels (1, 2, 3, 4, 5, 6, 7, or 8).

If you would like, I can also try to provide a patch to revert to the pre-d81614ca9068959884a6db3b78b8892e76055956 logic, though that's harder to test.

### DIFF
--- a/src/flac.c
+++ b/src/flac.c
@@ -43,6 +43,15 @@
 
 #define ENC_BUFFER_SIZE 8192
 
+/*
+** READ_LOOP_MAX_LEN is the maximum 'len' that will be passed to
+** flac_read_loop().  This is somewhat arbitrary, but must be less
+** than (UINT_MAX - FLAC__MAX_CHANNELS * FLAC__MAX_BLOCK_SIZE) to
+** avoid overflows, and must also be a multiple of the number of
+** channels (which is between 1 and 8.)
+*/
+#define READ_LOOP_MAX_LEN (0x10000 * 3 * 5 * 7)
+
 typedef enum
 {	PFLAC_PCM_SHORT = 50,
 	PFLAC_PCM_INT = 51,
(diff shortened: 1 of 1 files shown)


evpobr:
Hi @bemoody , thanks for pull request!

evpobr:
@arthurt ?

evpobr (approved):

arthurt (approved):
LGTM

evpobr:
Hi @arthurt !. @bemoody , thanks for contribution!

Issue #866, opened 2022-09-07
test suite fails with "rounding errors" for many non-x86 architectures


umlaeute:
i finally found the time to update the Debian package of `libsndfile` to 1.1.0, and after uploading to our build farm, the results are sobering.

while the build succeeds, the test suite does not for 12 out of 19 architectures.
of the 7 architectures that can successfully run the testsuite, 4 use a variant of the x86 CPU (x86, x86_64).

the failing architectures include `armhf` (of RaspberryPi fame) and `arm64` (people call it "the new thing").

here's a short tabular overview:

|arch | results |
|---|---|
|  [amd64](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=amd64&ver=1.1.0-1&stamp=1662500686&)          |  :tada:  |
|  [arm64](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=arm64&ver=1.1.0-1&stamp=1662500749&)          |  :boom:  |
|  [armel](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=armel&ver=1.1.0-1&stamp=1662501268&)          |  :boom:  |
|  [armhf](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=armhf&ver=1.1.0-1&stamp=1662500489&)          |  :boom:  |
|  [i386](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=i386&ver=1.1.0-1&stamp=1662500509&)            |  :tada:  |
|  [mips64el](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=mips64el&ver=1.1.0-1&stamp=1662500816&)    |  :boom:  |
|  [mipsel](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=mipsel&ver=1.1.0-1&stamp=1662501131&)        |  :boom:  |
|  [ppc64el](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=ppc64el&ver=1.1.0-1&stamp=1662500485&)      |  :boom:  |
|  [s390x](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=s390x&ver=1.1.0-1&stamp=1662512274&)          |  :boom:  |
|  [hppa](https://buildd.debian.org/status/fetch.php?pkg=libsndfile&arch=hppa&ver=1.1.0-1&stamp=1662502551&)            |  :tada:  |
[...]

Issue #860, opened 2022-08-03
MPEG stereo seek test fails on some architectures


mhlavink:
libsndfile 1.1.0 with enabled mpeg support fails on some architectures:
aarch64, ppc64le, s390x
works fine on x86_64 and i686
```
    mpeg_short_test                : mpeg_short.mp3 .......... ok
    mpeg_seek_test                 : mpeg_short.mp3 .......... ok
    mpeg_int_test                  : mpeg_int.mp3 ............ ok
    mpeg_seek_test                 : mpeg_int.mp3 ............ ok
    mpeg_float_test                : mpeg_float.mp3 .......... ok
    mpeg_seek_test                 : mpeg_float.mp3 .......... ok
    mpeg_double_test               : mpeg_double.mp3 ......... ok
    mpeg_seek_test                 : mpeg_double.mp3 ......... ok
    mpeg_stereo_seek_test          : mpeg_seek.mp3 ........... 
Line 306 : Error at index 0, got -0.000935562, should be -0.000935562. ### ppc64le
FAIL: tests/pedantic-header-test.sh

Line 306 : Error at index 0, got -0.000935562, should be -0.000935562. ### s390x
Line 324 : Error at index 4, got  1.70062e-05, should be  1.70062e-05. ### aarch64 - note different line, so first part passed
```

The values are just almost the same, but different. They start to differ on second to least significant bit iirc

buildroot environment:
gcc 12.1
lame 3.100
mpg123 1.30.1

evpobr:
Hi @mhlavink , thanks for report. It looks like a test flaw to me.

PR #867, opened 2022-09-08
tests: Use fuzzy comparison in test-suite

umlaeute:
Using exact comparison (e.g. `a == b`) when comparing expected with computed test data fails the test-suite on many architectures (including, but not limited to, `armhf` and `arm64`).

Instead, use epsilon(for now, `FLT_EPSILON` and `DBL_EPSILON`) to compare floating point numbers for equality.

Closes: https://github.com/libsndfile/libsndfile/issues/866

### DIFF
--- a/tests/utils.tpl
+++ b/tests/utils.tpl
@@ -193,6 +193,7 @@ sf_count_t		file_length_fd (int fd) ;
 #include <string.h>
 #include <ctype.h>
 #include <math.h>
+#include <float.h>
 #include <fcntl.h>
 #include <sys/stat.h>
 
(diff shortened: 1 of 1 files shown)


evpobr (approved):

umlaeute:
just a quick note: so far, the proposed changes have been tested on amd64 and arm64 only.
the test runs for the other (so-far failing) architectures, are still pending.

i can drop a note, once they have succeeded (or whether they have failed, and this PR was not sufficient to fix #866)

evpobr:
> can drop a note, once they have succeeded (or whether they have failed, and this PR was not sufficient to fix #866)

Yes please.

umlaeute:
as can be seen on the current [build status page](https://buildd.debian.org/status/package.php?p=libsndfile), the builds now all succeed.

at the time of writing, `sh4` and `m68k` are still building, but these archs already succeeded in the previous run (so I'm confident that they will pass this time as well)

SoapGentoo (approved):

evpobr:
Ok. maybe @arthurt well answer soon .

arthurt (approved):
Awesome stuff!

umlaeute:
btw,i figure that this patch is only used for the autotools-build, i don't know about the CMake-build (which I don't use).

evpobr:
Related #860.

evpobr:
Thank you all for participating!

Issue #868, opened 2022-09-11
public API header sndfile.h unneccessarily depends on Autoconf/CMake variables


manxorist:
All Autoconf/CMake variable shenanigans concerning `sf_count_t` are completely unneeded:
 *  `TYPEOF_SF_COUNT_T` is always `int64_t`, except for `WIN32` where it is `__int64`, which is ultimately an equivalent type. Even on Windows, `int64_t` is always available for all compilers as `sndfile.h` already includes `<stdint.h>` unconditionally anyway since 9 years ago (see <https://github.com/libsndfile/libsndfile/commit/88b1e0fad808808868dbd98d932b6e1a2b9bccdd>).
 *  `SIZEOF_SF_COUNT_T` is thus always `8`.
 *  and `SF_COUNT_MAX` is always `0x7fffffffffffffffLL`.
 *  `src/common.c` already fails to compile for `SIZEOF_SF_COUNT_T` != `8` anyway (see <https://github.com/libsndfile/libsndfile/blob/cefd7b59df628eca240af3c136d66137c8e94888/src/common.c#L570>).

Removing the configure/CMake trickery will make `sndfile.h` independent of build system feature detection, which is a huge simplification for people who want to build libsndfile with other build systems.

Pull request will follow.

PR #869, opened 2022-09-11
Make sndfile.h independent of build system variables

manxorist:
Fixes <https://github.com/libsndfile/libsndfile/issues/868>.


SoapGentoo:
Thanks, I wanted to get around to doing this myself too. The only downside (not for me!) is that some people will cry that their that MSVC 2008 support is lost with this. At this point, I wouldn't care about MSVC 2008 anymore, and people who want to use it should just add some macro hacks themselves.

SoapGentoo (approved):

manxorist:
>  The only downside (not for me!) is that some people will cry that their that MSVC 2008 support is lost with this.

As I noted in #868, VS2008 (which does not provide <stdint.h>) was already broken 9 years ago. I guess people would have complained by now. So this should really be a non-issue.

manxorist:
I'll look into the CI failures and update the pull request once I have figured it out.

SoapGentoo:
right, I didn't connect the dots. You will likely just have to fix that pedantic header test

evpobr:
Hi @manxorist .

evpobr:
@SoapGentoo , will you take this PR?

> The only downside (not for me!) is that some people will cry that their MSVC 2008 support is dropped with this. At this point, I wouldn't care about MSVC 2008 anymore, and people who want to use it should just add some macro hacks themselves.

We require C99-compliant compiler in `README.md`.

docs/index.md
```diff
@@ -138,7 +138,7 @@ Here is the release history for libsndfile:
   Linux platforms. Other numerous fixes to CMake build system, consider it
   stable now. Fix some memory leaks. Fix handling of some SD2 files. Update
   documentation. Integrate GitHub Actions for faster test builds and Oss-Fuzz
-  for fuzzing tests. Move sndfile.h.in from src/ to include/ directory. To avoid
+  for fuzzing tests. Move sndfile.h from src/ to include/ directory. To avoid
```
evpobr, line 141: Hmm, it is changelog of 1.0.30, do we really need to rewrite it?

evpobr (approved):

evpobr:
Thanks!

include/sndfile.h
```diff
@@ -365,9 +365,9 @@ typedef	struct sf_private_tag	SNDFILE ;
 ** and the Microsoft compiler.
 */
 
-typedef @TYPEOF_SF_COUNT_T@	sf_count_t ;
+typedef int64_t			sf_count_t ;
 #ifndef SF_COUNT_MAX
-#define SF_COUNT_MAX		@SF_COUNT_MAX@
+#define SF_COUNT_MAX		0x7FFFFFFFFFFFFFFFLL
```
SoapGentoo, line 370: @manxorist I thought a bit more, what about `INT64_MAX` here? That avoids the `long` vs `long long` literal discrepancy

manxorist: Yeah, that's probably a good additional cleanup. `LL`/`ll` suffix was what was always used before, I did just not consider changing it.

Do you want me to submit another pull request or will you just commit that yourself?

SoapGentoo: I'm a bit busy, so if you did it, that would help me :slightly_smiling_face:

manxorist: I'm also busy right now, I'll do it tomorrow. 🙂

manxorist: found some spare time, see #870

PR #870, opened 2022-09-12
Define SF_COUNT_MAX as INT64_MAX.

manxorist:
INT64_MAX instead of 0x7FFFFFFFFFFFFFFFLL avoids type discrepancy between int64_t and LL (long long) suffix when int64_t is not long long.


manxorist:
see https://github.com/libsndfile/libsndfile/pull/869#discussion_r968533920

SoapGentoo (approved):

PR #871, opened 2022-09-13
sndfile-play: use sndio when available

neheb:
libsndio is also available on Linux. Use it when found instead of alsa.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/programs/sndfile-play.c
+++ b/programs/sndfile-play.c
@@ -47,7 +47,10 @@
 
 #include "common.h"
 
-#if HAVE_ALSA_ASOUNDLIB_H
+#if HAVE_SNDIO_H
+	#include <sndio.h>
+
+#elif HAVE_ALSA_ASOUNDLIB_H
 	#define ALSA_PCM_NEW_HW_PARAMS_API
 	#define ALSA_PCM_NEW_SW_PARAMS_API
 	#include <alsa/asoundlib.h>
@@ -60,9 +63,6 @@
 	#include 	<fcntl.h>
 	#include 	<sys/ioctl.h>
 	#include 	<sys/soundcard.h>
-
-#elif HAVE_SNDIO_H
-	#include <sndio.h>
 
 #elif (defined (sun) && defined (unix)) || defined(__NetBSD__)
 	#include <fcntl.h>


SoapGentoo:
I'm not sure that changing to sndio over ALSA by default is the right decision, _especially_ on Linux. If you want to use sndio, you should 1. add a configure option 2. change the checks in `configure.ac` 3. disable alsa support. This default behaviour change is not going to fly.

evpobr:
Hi @neheb . Yes, as mentioned above, this will break the behavior of the utility for many other people. Use the options to configure.

SoapGentoo:
Finally, I see that `sndio` has a `sndio.pc` file, hence requesting it should be done via `pkg-config`.

neheb:
hmm I'm unsure how to change CMake to reflect this.

evpobr:
It seems like we check Sndio presence on OpenBSD platform only.

@SoapGentoo , does it make sense to check it also on Linux platform?

SoapGentoo:
yes, we package `sndio` in Gentoo, and it also works fine on Linux, but
1. it definitely shouldn't be the default
2. and it will be mutually exclusive with ALSA
3. I don't like how it's gathering its _CFLAGS/_LIBS right now, i.e. it should use `PKG_CHECK_MODULES`

evpobr:
Yes, in our code ALSA is default and exclusive with Sndio.

janstary:
Actually, this bit of cmake/SndFileChecks.cmake

```
if (CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
        find_package (Sndio)
elseif (NOT WIN32)
        find_package (ALSA)
endif ()
```

entirely ignores anything but Linux, with the exception of OpenBSD:
if on OpenBSD, look for sndio; otherwise, you surely have ALSA.
Not of macOS, for example.

brad0:
> It seems like we check Sndio presence on OpenBSD platform only.
> 
> @SoapGentoo , does it make sense to check it also on Linux platform?

Linux and FreeBSD. and yes, as an OpenBSD developer that typically contributes these kinds of diffs, please switch to using the sndio pkg-config file.

evpobr:
@SoapGentoo , i will be very buzy this month. Can you take PRs?

SoapGentoo:
@brad0 agreed
@neheb please turn sndio into a full-fledged option. You'll likely have to turn it into `--with-sound-backend={sndio,alsa}` option, seeing they're generally mutually exclusive

neheb:
not quite sure how to.

SoapGentoo:
then I have to do it, which means you'll need to wait another week or so

janstary:
> Finally, I see that `sndio` has a `sndio.pc` file, hence requesting it should be done via `pkg-config`.

I don't see any `*.pc` file in either [the release tarball](https://sndio.org/sndio-1.9.0.tar.gz) or
the git repo at  https://caoua.org/git/sndio

SoapGentoo:
```
install:
...
		cp sndio.pc ${DESTDIR}${PKGCONF_DIR}
```
in `libsndio/Makefile.in`

janstary:
Ah, fat fingers, sorry .

PR #881, opened 2022-10-03
mpeg_decode: Fix bug where final bytes from mpg123_read() where ignored.

### LINKED ISSUE #880: libsndfile (soundfile) for mp3 not float32 but float64
Link to issue [libsndfile (soundfile) for mp3](https://github.com/bastibe/python-soundfile/issues/349)
Audio out from libsndfile (soundfile) for mp3 not float32 but float64.
Because of this, if we do not force the dtype=float64 , we get an empty array

print mp3 with dtype set float32 by default
audio_test, _ = librosa.load('./g.mp3', mono=False, res_type='kaiser_fast',sr=sr)
```
[]
0
(2, 0)
tensor([], size=(2, 0))
```
print mp3 with dtype =float64
audio_test, _ = librosa.load('./g.mp3', mono=False, res_type='kaiser_fast',sr=sr, dtype=np.float64)
```
[[0. 0. 0. ... 0. 0. 0.]
 [0. 0. 0. ... 0. 0. 0.]]
21182464
(2, 10591232)
tensor([[0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.]])
```

arthurt:
Resolve #880 

Most every read api I've ever worked with return no error and positive byte counts, or zero bytes and an end or error condition. When the end of stream is reached, the last read will contain the last bytes, and the next call to read returns an error indicating that end has been reached.

But not mpg123, no, when reaching the last bytes, it returns the last bytes AND sets the error condition to done.

Assumptions make and ass of you an mumptions.

### DIFF
--- a/src/mpeg_decode.c
+++ b/src/mpeg_decode.c
@@ -91,11 +91,8 @@ mpeg_dec_decode (SF_PRIVATE *psf, float *ptr, sf_count_t len)
 
 	error = mpg123_read (pmp3d->pmh, (unsigned char *) ptr, len * sizeof (float), &done) ;
 
-	if (error == MPG123_OK)
+	if (error == MPG123_OK || error == MPG123_DONE)
 		return done / sizeof (float) ;
-
-	if (error == MPG123_DONE)
-		return 0 ;
 
 	if (error == MPG123_NEW_FORMAT)
 	{	psf->error = SFE_MALFORMED_FILE ;


evpobr (approved):

evpobr:
Thanks @arthurt !

SoapGentoo (approved):

Issue #882, opened 2022-10-06
CPUs might clip integers but not shorts


umlaeute:
recently the [`python-soundfile`](https://github.com/bastibe/python-soundfile) wrapper for `libsndfile` failed one of its unit-tests: writing a large negative number (e.g. `-2.0^15 - 1 == 32769.0`) into a `float`-soundfile and reading it back as `short`.
the unit-test only failed on 32bit machines.

the full bug report can be found at https://github.com/bastibe/python-soundfile/issues/350

after a lot of digging, it turns out that libsndfile does not properly clip values, even if [SFC_SET_CLIPPING](http://www.mega-nerd.com/libsndfile/command.html#SFC_SET_CLIPPING) is enabled.

this is because libsndfile tries to utilize automatic CPU-based clipping if possible, and only manually clips the values if it finds that the CPU cannot do so, see the `CPU_CLIPS_*` checks in https://github.com/libsndfile/libsndfile/blob/0faf893fd21866bf08a5e64adf2ccf8d7ab0a9de/src/float32.c#L450-L452

CPU-clipping support is detected in [m4/clip_mode.m4](https://github.com/libsndfile/libsndfile/blob/0faf893fd21866bf08a5e64adf2ccf8d7ab0a9de/m4/clip_mode.m4), which tests if floats are clipped when converted to 32bit integers.

now the fun part is this: the 32bit machines I ran this on ("i386" and "armhf" Debian architectures), actually support CPU-based clipping, but apparently *only* for 32bit integers, and **not** for 16bit `short`.

Unfortunately, such subtleties are not taken into account when libsndfile attempts to clip values, and as a consequence `f2s_clip_array ()` will not clip on 32bit architectures.

It's possible to disable the clipping test entirely, but I would prefer it if libsndfile would do the right thing from the start.

umlaeute:
i found that of the 10 different 64bit platforms supported by Debian, not a single one would support CPU-based clipping (as tested by `m4/clip_mode.m4`.
OTOH, all 32bit platforms had some sort of CPU-based clipping support (either for negative numbers (x86) or for both positive and negative numbers (another 8 architectures supported by Debian).

given the prevalence of 64bit machines and that all relevant 32bit machines (i386 and arm; i'm not entirely sure how committed you are to have performance tweaks for m68000 CPUs) wrongly assume that they can do CPU-based clipping of `short`s, i think it would be best to just unconditionally do manually clipping in `f2s_clip_array ()`

[...]

PR #883, opened 2022-10-06
Fix clipping of non-int32 numbers

evpobr:


SoapGentoo (approved):
give a rationale why we can remove the conditional macros now?

evpobr:
@umlaeute explained everything well #882.

umlaeute:
@arthurt i haven't checked whether this PR removes *all* uses of `CPU_CLIPS_*`, or only those that are not about `int32_t`.
if the latter, than the `--disable-cpu-clip` flag still makes sense...

umlaeute:
as a side note: you might also want to add some unit-tests for correct clipping behaviour.

it was by mere chance that we triggered the failing test in the `python-soundfile` package on our Debian infrastructure, as the package currently only runs the testsuite as part of the build process. and since `python-soundfile` itself is architecture independent (with all the architecture-specific parts in its dependencies, like e.g. `libsndfile`), we usually don't build the package on any architecture except for the default one (which is x86_64).
in contrast, the `libsndfile` testsuite is run on all architectures that can build the library.

it would have also helped (and saved me a lot of time), if I could have been certain from the beginning, in which component the issue actually originated. (finding bugs is hard enough in a software package itself; but if you have to chase down dependencies, it becomes truly awful)

evpobr:
> @arthurt i haven't checked whether this PR removes _all_ uses of `CPU_CLIPS_*`, or only those that are not about `int32_t`. if the latter, than the `--disable-cpu-clip` flag still makes sense...

I applied your patch without changes:

https://salsa.debian.org/multimedia-team/libsndfile/-/blob/master/debian/patches/short_clipping.patch

evpobr:
Let's merge this pull request, it's definitely useful. Then you can think about the details. @arthurt ?

PR #884, opened 2022-10-06
wavlike: Reorganize wavlike_subchunk_parse(), Check for More Error Conditions

### LINKED ISSUE #873: WAV: Check chunk boundary while parsing LIST subchunks
### Discussed in https://github.com/libsndfile/libsndfile/discussions/872

LIST subchunks may include padding so that the chunk is aligned to 4-byte offsets. We need to check that there are enough bytes left in the chunk before reading the next list member marker and size. Otherwise, we overshot the chunk, and upon return from parsing the subchunk, the header parser location may be out-of-sync.

This issue is in `wavlike_subchunk_parse ()`

arthurt:
Reorganize the loop within `wavlike_subchunk_parse()`.

Fix the chunk-remaining space checks to actually check for enough space for the next bytes read.

Handle short-reads errors correctly.

Don't error if unknown sub-chunks are larger than the not-used buffer..

Fixes #873 .

### DIFF
--- a/src/wavlike.c
+++ b/src/wavlike.c
@@ -952,7 +952,7 @@ int
 wavlike_subchunk_parse (SF_PRIVATE *psf, int chunk, uint32_t chunk_length)
 {	sf_count_t	current_pos ;
 	char		buffer [2048] ;
-	uint32_t	chunk_size, data_size, bytesread = 0 ;
+	uint32_t	chunk_size, bytesread = 0 ;
 
 	current_pos = psf_fseek (psf, 0, SEEK_CUR) ;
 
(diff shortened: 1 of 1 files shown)


evpobr:
CI Fuzz test unexpectedly returned an error.

arthurt:
Uggg, The fuzzing setup isn't what I would call, easily repeatable.

arthurt:
Fuzzer passed, but still giving warnings...

evpobr:
Yes, I checked the previous run, so there were no such warnings.

cmeister2:
@evpobr @arthurt I've run the fuzzer locally using the oss-fuzz reproduce tool and the crash from the CI pipelines and it doesn't reproduce for me. Might be worth just retrying the pipeline to see if it was a one-off failure or not. I'll keep poking it a bit more though.

evpobr:
Ok, restarted.

kiilerix:
It seems like this PR had some good cleanup ... which however was hard to review and conflicted with 890. I hope it will make a comeback.

Issue #374, opened 2018-03-16
Error parsing/loading .wav files recorded with a Zoom R8


x42:
This is a follow-up of #370 

A file recorded with a Zoom, anonymized by zeroing audio-data after 0x00060010 (audio-data starts at 0x00060000): http://robin.linuxaudio.org/tmp/zoom_sndfile_bug.wav.gz  

with libsndfile-1.0.27:

```
File : zoom_sndfile_bug.wav
Length : 27293176
RIFF : 27293168
WAVE
fmt  : 16
  Format        : 0x1 => WAVE_FORMAT_PCM
  Channels      : 1
  Sample Rate   : 44100
  Block Align   : 2
  Bit Width     : 16
  Bytes/sec     : 88200
LIST : 393164 (too long)
data : 26899960
End

----------------------------------------
Sample Rate : 44100
Frames      : 13449980
Channels    : 1
Format      : 0x00010002
Sections    : 1
Seekable    : TRUE
Duration    : 00:05:04.988
Signal Max  : 42 (-57.84 dB)
```

With libsndfile  1.0.28-115-g65eabcbf (1.0.28 release is also affected):
```
Error : Not able to open input file /tmp/zoom_sndfile_bug.wav.
File : /tmp/zoom_sndfile_bug.wav
Length : 27293176
RIFF : 27293168
WAVE
fmt  : 16
  Format        : 0x1 => WAVE_FORMAT_PCM
  Channels      : 1
  Sample Rate   : 44100
  Block Align   : 2
  Bit Width     : 16
  Bytes/sec     : 88200
LIST : 393164
    *** ��N
            : 50466568
Request for header allocation of 786320 denied.
*** Unknown chunk marker (19A75F8) at position 393212. Exiting parser.
*** Chunk size 4292673536 > file length 27293176. Exiting parser.

Error in WAV file. No 'data' chunk marker.
```

PS. the output `*** ��N` comes from the "%M" in src/wavlike.c:1035, an  unhandled default chunk.

PPS. The file plays fine with mplayer, ffmpeg and also older tools using libsndfile1.0.27

sciurius:
What Zoom device did make this?

x42:
Zoom R8

mxmilkiib:
Btw, this is blocking Ardour from using 1.0.28 and above.

PR #890, opened 2022-10-26
Let wav parsing recover from invalid LIST chunks, fixing #374

kiilerix:
This triggered other minor fixes and cleanup. See the individual commits.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -375,7 +375,7 @@ psf_log_printf (SF_PRIVATE *psf, const char *format, ...)
 					strptr = istr ;
 					while (*strptr)
 					{	c = *strptr++ ;
-						log_putchar (psf, c) ;
+						log_putchar (psf, psf_isprint (c) ? c : '.') ;
 						} ;
 					break ;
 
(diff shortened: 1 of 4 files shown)


evpobr:
Hi @kiilerix , impressive work! @arthurt , can you take a look?

arthurt (approved):

arthurt:
It's less invasive that #884 .

kiilerix:
> It's less invasive than #884 .

I wasn't aware of that one ... or whether it solves the same problem. It seems to have some nice things. And some that seems more debatable. But too much happening at once makes it hard to tell, so IMO that one should be reworked anyway ... hopefully after being rebased on top of this one.

arthurt:
:+1: I think #884 should be deferred for your PR.

evpobr (approved):

evpobr:
Thanks @kiilerix !

evpobr:
I suggest accepting this pull request if it helps right now. @arthurt ?

evpobr:
Ah, you approved. Ok then.

Issue #830, opened 2022-04-29
Test suite failure on x86_64-musl


lun-4:
Hello! I've been trying to update libsndfile from 1.0.31 to 1.1.0 in Void Linux, and I found this failure as its CI attempted to compile and test it for an x86_64 system using the musl libc. At the first time I upgraded, the test suite failed due to missing `libmpg123` and `liblame` (it was saying it wasn't compiled with MP3 support, which makes sense), after adding them both, I find the following error:

Reference for the update patch itself: https://github.com/void-linux/void-packages/pull/36908

Reference for the failing test suite: https://github.com/void-linux/void-packages/runs/6222001344?check_suite_focus=true

```
    pcm_test_int                   : wrt_32bit.sd2 ........... Note: Illegal Audio-MPEG-Header 0x00001831 at offset 9140.
Note: Trying to resync...
Note: Skipped 4 bytes in input.
Note: Illegal Audio-MPEG-Header 0x00005280 at offset 9148.
Note: Trying to resync...
Note: Skipped 28 bytes in input.
Note: Illegal Audio-MPEG-Header 0x00002207 at offset 9188.
Note: Trying to resync...
Note: Skipped 4 bytes in input.
Note: Illegal Audio-MPEG-Header 0x0000180b at offset 9196.
Note: Trying to resync...
Note: Skipped 16 bytes in input.
Note: Illegal Audio-MPEG-Header 0xffffeda8 at offset 9216.
Note: Trying to resync...
Note: Skipped 1 bytes in input.


Line 4535 : Stereo : sf_seek (file, 7091, SEEK_SET) returned 20915 (should be 7091).
```

I don't use the musl libc myself, but does that error make sense with a libc change?

evpobr:
@lun-4 .

@arthurt , can you look at this please?

arthurt:
When a test fails, the audio fail used in the test is not deleted. Could you check your working directory for a *.mp3 file that the failed test created and attach it here? Thanks!

lun-4:
There isn't a `.mp3` file inside the build directory, but here's `wrt_32bit.sd2`, since that's the file being referenced in the failing test. file(1) can't identify it as an mp3 either, it's just "data".

[libsndfile_failing_test.zip](https://github.com/libsndfile/libsndfile/files/8591931/libsndfile_failing_test.zip)

arthurt:
Okay, I have a theory.

[...]

PR #898, opened 2022-12-12
Identify 'naked' MPEG streams last, even after resource forks

arthurt:
The minimal possible MPEG file contains no headers or identification other than the brief sync header. This header is only 3 bytes, and is quite prone to false-positives. Particularly raw PCM can look like a sync header.

As such, move detection of 'naked' MPEG files in `guess_file_type()` to the very last test. Give more weight to a sync header if it follows an ID3 tag.

In issue #830, differences in libc random number functions mean that SD2 files generated by the test suite looked like mp3 files. SD2 files are raw data with the header stored in a resource fork. This PR move the test for 'naked' MPEG to after even checking for resource forks.

It may be preferable to remove the check for naked MPEG streams altogether, as if `guess_file_type()` can't return a match, `format_from_extension()` is tried, which includes checks for `.mp3`. Using the extension is arguably a more reliable method in this case.

### DIFF
--- a/src/sndfile.c
+++ b/src/sndfile.c
@@ -2772,6 +2772,17 @@ format_from_extension (SF_PRIVATE *psf)
 } /* format_from_extension */
 
 static int
+identify_mpeg (uint32_t marker)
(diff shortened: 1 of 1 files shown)


evpobr:
@SoapGentoo ?

evpobr (approved):

SoapGentoo (approved):

PR #899, opened 2022-12-12
wavlike: Find a use for unused bytes total

arthurt:
Compile upgrades have found an existing set-but-unused variable in `wavlike_read_cart_chunk()`

Rather than remove the unread `bytes` variable (which possibly points to another bug, that read's don't appear to be bounded), use the bytes tally for a debug message about padding bytes.


evpobr:
@arthurt , thanks for fixes. Now we ready for 1.2.0?

evpobr (approved):

SoapGentoo (approved):

PR #904, opened 2023-01-13
get rid of html remnants

janstary:
Now that the docs are in *.md, the 'black-on-white' html option etc can be removed.
There is still a few html files under docs/ - is that what http://libsndfile.github.io/libsndfile/ is built from?


evpobr (approved):

evpobr:
> There is still a few html files under docs/ - is that what http://libsndfile.github.io/libsndfile/ is built from?

In `_includes` & `_layout` directories? Yes, it is required for [site](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll).

SoapGentoo:
@evpobr I haven't checked out the site works, hence don't have an opinion here

PR #908, opened 2023-01-15
Handle absolute CMAKE_INSTALL_LIBDIR/INCLUDEDIR

Staudey:
A fellow packager for Solus discovered that the CMakeLists rules currently expect a relative path (without prefix) for `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` when setting the variables for the pkgconfig. AIUI it's valid for those paths to be absolute (with prefix), which causes the prefix to be prepended twice. So in the case of our packaging scripts this resulted in `libdir=${prefix}//usr/lib64` (`/usr` being the prefix).

This PR adds logic inspired by other projects to only prepend the prefix if the path isn't already absolute.


SoapGentoo:
I was aware of the issue, but imo this is the wrong approach. You should just put `CMAKE_INSTALL_FULL_LIBDIR` into the pkgconfig files.

Staudey:
That would certainly work for our purposes too. It certainly would shorten the whole thing. I'm not sure which edge cases the logic I used here was guarding against in other projects, but I see no issues with using the `FULL_LIBDIR` from my naive point of view. Will update accordingly.

SoapGentoo:
The CMake build system is literally a 1:1 bug-compatible translation from autoconf, and hence copied the prefix-handling too. That said, we generally recommend using the autotools build system on POSIX systems still. Any reason not to use it?

Staudey:
@SoapGentoo I think I simply had the false impression that the CMake system was now preferable (possibly by several CMake changes mentioned in the latest release notes), and didn't see that you still recommended autotools in your README, my bad.

evpobr:
It was not ideal, that's true. Thanks @Staudey .

PR #926, opened 2023-04-10
fix #888, change diffrence of 'ret', 'offset' form Xiph.org's Opusfile: opusfile-0.12/src/opusfile.c

### LINKED ISSUE #888: Failed to read some OPUS audio
we have a lot of OPUS audio, most of them can be read well, some got failed, but the failed OPUS can be played by Windows media player and mplayer.
The failed example is here: https://drive.google.com/file/d/1_3oFZl7Ruv64rAMRCWCpTR1QkAEtZJSo/view?usp=sharing

Using `sndfile-info` to check the audio, it can't get the `Frames` and `Duration`:
```
========================================
File : zz.opus
Length : 118784
Ogg stream data : Opus
Stream serialno : 409619929
Opus library version: libopus 1.3.1
Opus Header Metadata
  OggOpus version  : 1
  Channels         : 1
  Preskip          : 312 samples @48kHz
  Input Samplerate : 16000 Hz
  Gain             : 0.0
  Channel Mapping  : 0 (mono or stereo)
VorbisComment Metadata
  Vendor: libopus 1.1
  ENCODER=opusenc from opus-tools 1.1
  Granule pos offset  : 0

----------------------------------------
Sample Rate : 16000
Frames      : unknown
Channels    : 1
Format      : 0x00200064
Sections    : 1
Seekable    : TRUE
Duration    : unknown
```

veelion:
I compared the relative part of `ogg.c` with Xiph.org's opusfile-0.12/src/opusfile.c, and found there is a little difference between them. changing it makes #888 succeed.

### DIFF
--- a/src/ogg.c
+++ b/src/ogg.c
@@ -424,12 +424,13 @@ ogg_sync_last_page_before (SF_PRIVATE *psf, OGG_PRIVATE *odata, uint64_t *gp_out
 		left_link = 0 ;
 		while (position < end)
 		{	ret = ogg_sync_next_page (psf, &odata->opage, end - position, &position) ;
-			if (ret <= 0)
+			if (ret < 0)
 				return -1 ;
+			else if (ret == 0) break ;
 			if (ogg_page_serialno (&odata->opage) == serialno)
 			{	uint64_t page_gp = ogg_page_granulepos (&odata->opage) ;
 				if (page_gp != (uint64_t) -1)
-				{	offset = position ;
+				{	offset = ret ;
 					gp = page_gp ;
 					}
 				}
--- a/src/ogg.c
+++ b/src/ogg.c
@@ -430,7 +430,7 @@ ogg_sync_last_page_before (SF_PRIVATE *psf, OGG_PRIVATE *odata, uint64_t *gp_out
 			if (ogg_page_serialno (&odata->opage) == serialno)
 			{	uint64_t page_gp = ogg_page_granulepos (&odata->opage) ;
 				if (page_gp != (uint64_t) -1)
-				{	offset = ret ;
+				{	offset = position ;
 					gp = page_gp ;
 					}
 				}


evpobr:
Hi @veelion . @arthurt ?

evpobr:
Hi @veelion , we have build errors.

veelion:
> Hi @veelion , we have build errors.

 `make` succeed, I found the errors came from `make distcheck`, I'll check it.

veelion:
Hi @evpobr , I've checked the build log,  and checked the code without my PR, it also has error when `make distcheck`. I think the error is not from my PR
for `build (ubuntu-gcc-autotools) `, the error is about read `.oga`, not `.opus`
```
Line 264: sf_open (SFM_READ) failed : Unspecified internal error.

File : fpt_vorbis.oga
Length : 3947
Ogg stream data : Vorbis
Stream serialno : 1510130196
Vorbis library version : Xiph.Org libVorbis 1.3.7
Bitstream is 1 channel, 16000 Hz
Encoded by : Xiph.Org libVorbis I 20200704 (Reducing Environment)
Ogg: Junk after the last page.
PCM offset  : 0
PCM end     : 8192
End
Invalid SF_PRIVATE field : datalength == -2444.

FAIL: tests/pedantic-header-test.sh
======================================
1 of 1 test failed
Please report to sndfile@mega-nerd.com
make[4]: *** [Makefile:3375: check-TESTS] Error 1
```

for `build (ubuntu-gcc-cmake)`, the errors are: 
```
The following tests FAILED:
	  8 - floating_point_test (Failed)
	130 - ogg_test (Failed)
	131 - compression_size_test_vorbis (Failed)
	133 - string_test_ogg (Failed)
	134 - misc_test_ogg (Failed), 
```
but about opus is passed:
```
        Start 135: ogg_opus_test
135/143 Test #135: ogg_opus_test ....................   Passed    0.03 sec
```
My PR is only about opus, but the errors above is about others. What do you think?

arthurt:
The routines in `src/ogg.c` are used by both Ogg/Vorbis and Ogg/Opus (`.opus`), subformats. So breakage of the Ogg/Vorbis tests is very likely a result of this change.

While yes, a lot of the routines in ogg.c were copied from opusfile, they were also somewhat modified, as libsndfile doesn't support all the Ogg features in Opusfile. We'll have to dig more into why the sample Opus files don't open correctly.

arthurt:
Looking at the file from #888, it appears that the EOS marker is not set on the last page of the stream.

Technically not a malformed file, as Ogg does allow for EOF without an EOS marker if there is no chaining, to allow for truncated streams, but certainly could be related to the issue.

veelion:
> The routines in `src/ogg.c` are used by both Ogg/Vorbis and Ogg/Opus (`.opus`), subformats. So breakage of the Ogg/Vorbis tests is very likely a result of this change.
> 
> While yes, a lot of the routines in ogg.c were copied from opusfile, they were also somewhat modified, as libsndfile doesn't support all the Ogg features in Opusfile. We'll have to dig more into why the sample Opus files don't open correctly.

I see why test of Ogg/Vorbis faied, I only thinked about opus, but it affected  Vorbis. yes, we have to dig more.

evpobr:
Looks good now. @arthurt ?

arthurt (approved):

evpobr:
Thanks @veelion and @arthurt!

PR #930, opened 2023-04-27
Clamp wav header filelength and datalength

magnus-nomono:
Libsndfile uses `int64_t` to represent datalength and filelength internally, but the wave/riff header is limited to `uint32`.  Ideally the user should stop writing to the file before reaching this max, but since the library allows writing the PCM data beyond this, it would in my opinion be better to clamp the header values instead of the wrap around we get today.

This is a minimal patch suggestion for wav, but might want to create a shared solution in `psf_binheader_writef` instead (or just disallow writing beyond the max, depending on preference)

### DIFF
--- a/src/wav.c
+++ b/src/wav.c
@@ -79,7 +79,7 @@
 
 #define WAVLIKE_PEAK_CHUNK_SIZE(ch) 	(2 * sizeof (int) + ch * (sizeof (float) + sizeof (int)))
 
-
+#define MIN(a,b) ((a) < (b) ? (a) : (b))
 enum
 {	HAVE_RIFF	= 1 << 0,
 	HAVE_WAVE	= 1 << 1,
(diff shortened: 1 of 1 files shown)


src/wav.c
```diff
@@ -79,7 +79,7 @@
 
 #define WAVLIKE_PEAK_CHUNK_SIZE(ch) 	(2 * sizeof (int) + ch * (sizeof (float) + sizeof (int)))
 
-
+#define MIN(a,b) ((a) < (b) ? (a) : (b))
```
arthurt, line 82: Please use the already defined `SF_MIN()` macro instead.

arthurt (changes requested):
Looks good, small change to avoid a duplicate macro definition.

arthurt:
Thanks for the contribution!

Yes, writing the overflow value is not a desired behaviour.

A WAV length value of 0xFFFFFFFF is interpretted as meaning 'unknown.' So a side effect of this change is that when a length value is too large to fit in the 32-bit field, it will appear as unknown, which is probably the best solution.

src/wav.c
magnus-nomono: Ah, was searching for an existing macro, but for some reason it didn't come up. Thanks, will fix

arthurt (approved):

evpobr (approved):

PR #939, opened 2023-05-27
Add move semantics for SndfileHandle if C++11 or newer

haydaralaidrus:
Adding move constructor and move assignment for `SndfileHandle` if the C++11 std or newer is supported.

### DIFF
--- a/include/sndfile.hh
+++ b/include/sndfile.hh
@@ -100,6 +100,11 @@ class SndfileHandle
 			SndfileHandle (const SndfileHandle &orig) ;
 			SndfileHandle & operator = (const SndfileHandle &rhs) ;
 
+#if (__cplusplus >= 201100L)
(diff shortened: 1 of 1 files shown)


evpobr:
Hi @haydaralaidrus .

evpobr (commented):
I'm not sure, in C++11 mode do we need to mark copy constructor and copy assignment operator as deleted?

haydaralaidrus:
Hi @evpobr !,

Do you mean in this particular case or in general? Im sorry if i don't follow your question. But, if you want to prevent the compiler from generating the default copy constructor and copy assignment, yes. Either delete them or define them ourselves. I don't see why it is necessary to delete it here, though.

evpobr:
Ok then.

evpobr:
Thanks @haydaralaidrus !

haydaralaidrus:
@evpobr no problem!

PR #940, opened 2023-05-30
Fix realloc use

Halmoni100:
Compiling with gcc 12.1 reveals that the pointer passed to realloc may be misused since it is potentially invalidated (compile fails).

Fix with clearer code that does compile.


arthurt (approved):
LGTM. Thank you for this necessary change!

evpobr:
Thanks @Halmoni100 !

PR #942, opened 2023-06-02
AIFF fixes

arthurt:
Two fixes related to the AIFF container.

## Drop PEAK info if channel count changes

Chunks inside of an AIFF file can come in any order. There are two places chunks which can specify channel count, the containing FORM chunk and the COMM chunk. The PEAK chunk doesn't specify a size, and is proportional to the known channel count. If the COMM chunk occurs after the PEAK chunk, and has a different channel count, the PEAK chunk should be discarded. Firstly because there is no reasonable behaviour on how to adapt the peak info, and secondly because the memory allocation for the peak info is no longer the correct size.

## Ignore data padding when re-writing header

AIFF files can have an arbitrary gap between the end of the header data and the start of the audio data. (This is referenced by the AIFF spec as to aid accelerated filesystem block aligned writes.) These extra pad bytes, if present, are not read by libsndfile. When rewriting the AIFF header in the case of a file opened for RDWR, previously entire range of [0, psf->dataoffset] was used as the header data. As `dataoffset` pointed to the first byte of audio data, it included these padding bytes.

Firstly, this is a waste of effort, as the pad bytes should be ignored and unchanged.

Secondly, as the binheader buffer mechanism is used to read the header, and the pad bytes were not read when initially parsing the header, the additional read of the pad bytes can potentially overflow the existing memory allocation of the binheader buffer. A maliciously crafted file could do this easily.

Instead, calculate the actual header size from the saved SSND chunk offset and fixed size of the SSND chunk before the padding bytes. Additionally, this PR adds some sanity checks before reusing the binheader buffer.

## Additional Changes

- Fix a style error in `common.c`
- ~~Move the maximum binheader allocation size to a `#define`~~
- ~~Allow for allocation up to the binheader maximum size.~~

Binheader changes stripped due to fuzzing issues for now.

### DIFF
--- a/src/aiff.c
+++ b/src/aiff.c
@@ -1163,18 +1163,27 @@ aiff_read_comm_chunk (SF_PRIVATE *psf, COMM_CHUNK *comm_fmt)
 /*==========================================================================================
 */
 
-static void
+static int
 aiff_rewrite_header (SF_PRIVATE *psf)
-{
+{	AIFF_PRIVATE *paiff = psf->container_data ;
+
 	/* Assuming here that the header has already been written and just
 	** needs to be corrected for new data length. That means that we
 	** only change the length fields of the FORM and SSND chunks ;
 	** everything else can be skipped over.
 	*/
 	int k, ch, comm_size, comm_frames ;
+	sf_count_t header_len ;
 
+	/* Calculate the header length rather than use dataoffset, as AIFF files
+	** can have additional padding offset bytes which aren't usefully a part of
+	** the header.
+	*/
+	header_len = paiff->ssnd_offset + 8 + SIZEOF_SSND_CHUNK ;
+	if (psf->header.len < header_len || header_len > psf->dataoffset)
+		return SFE_INTERNAL ;
 	psf_fseek (psf, 0, SEEK_SET) ;
-	psf_fread (psf->header.ptr, psf->dataoffset, 1, psf) ;
+	psf_fread (psf->header.ptr, header_len, 1, psf) ;
 
 	psf->header.indx = 0 ;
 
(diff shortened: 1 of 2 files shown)


src/aiff.c
```diff
@@ -1059,6 +1059,13 @@ aiff_read_comm_chunk (SF_PRIVATE *psf, COMM_CHUNK *comm_fmt)
 	else
 		psf_log_printf (psf, "  Sample Size : %d\n", comm_fmt->sampleSize) ;
 
+
+	if (psf->sf.channels != comm_fmt->numChannels && psf->peak_info)
```
SoapGentoo, line 1063:
```suggestion
	if ((psf->sf.channels != comm_fmt->numChannels &&) psf->peak_info)
```
I like not having to think about precedence of binary operations

src/aiff.c
```diff
@@ -1059,6 +1059,13 @@ aiff_read_comm_chunk (SF_PRIVATE *psf, COMM_CHUNK *comm_fmt)
...
+	{	psf_log_printf (psf, "  *** channel count changed, discarding existing PEAK chunk\n") ;
+		free (psf->peak_info) ;
+		psf->peak_info = NULL ;
+		} ;
```
SoapGentoo, line 1067:
```suggestion
		}
```

SoapGentoo (changes requested):

src/aiff.c
arthurt: That's actually a style thing. Erik's style was to put `;` if the `if` doesn't have an `else`, and it does occasionally catch indentation bugs, so I've kept it up.

src/aiff.c
arthurt: What about 
```c
	if ((psf->sf.channels != comm_fmt->numChannels) && psf->peak_info)
```
?

SoapGentoo: see, proving my case :D

src/aiff.c
SoapGentoo: oh ok, some compilers warn on this, hence my comment

src/aiff.c
SoapGentoo: yes, mine obviously can't work :wink:

SoapGentoo (approved):

evpobr (approved):

PR #943, opened 2023-06-17
mpeg: fix reading of mpeg files that lack an INFO header

JJL772:
I encountered some funky MP3 files that are not using the ID3 container, and cannot be successfully read by sndfile. The library correctly identifies them as mpeg files based on the frame header, but it thinks they have no data because mpg123_length returns 0. According to the [mpg123 docs](https://www.mpg123.de/api/group__mpg123__status.shtml#ga7a97295f5dd82795489d9ed8ee544805), length relies on either an info header at the start of the file or a previous call to mpg123_scan.

Not sure if this is the most optimal solution, but it works. Let me know if my commit message is inaccurate, I'm not an audio programmer :)

### DIFF
--- a/src/mpeg_decode.c
+++ b/src/mpeg_decode.c
@@ -203,7 +203,7 @@ mpeg_dec_seek (SF_PRIVATE *psf, int mode, sf_count_t count)
 } /* mpeg_dec_seek */
 
 static int
-mpeg_dec_fill_sfinfo (mpg123_handle *mh, SF_INFO *info)
+mpeg_dec_fill_sfinfo (SF_PRIVATE* psf, mpg123_handle *mh, SF_INFO *info)
 {	int error ;
 	int channels ;
 	int encoding ;
@@ -218,6 +218,12 @@ mpeg_dec_fill_sfinfo (mpg123_handle *mh, SF_INFO *info)
 	info->channels = channels ;
 
 	length = mpg123_length (mh) ;
+	if (length <= 0 && !psf->is_pipe)
+	{	if ((error = mpg123_scan(mh)) != MPG123_OK)
+			return error ;
+		length = mpg123_length(mh) ;
+		}
+
 	if (length >= 0)
 	{	info->frames = length ;
 		info->seekable = SF_TRUE ;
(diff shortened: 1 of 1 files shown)


arthurt:
Hi, and thanks for your contribution!

Small clarification. The header that mpg123 is referencing is the Xing or INFO header that some (most) MP3 encoders insert as part of the MPEG stream at the beginning, and is not related to ID3 tags.

ID3v2 does have the capability to store the stream length, but AFAIK mpg123 and most decoders ignore this, as usually the INFO header is present and more likely to be accurate.

I'm curious about the files which were unable to open. Is it possible to upload them? It would also help us validate the fix works.

Your changes generally look good :)

src/mpeg_decode.c
```diff
@@ -218,6 +218,11 @@ mpeg_dec_fill_sfinfo (mpg123_handle *mh, SF_INFO *info)
 	info->channels = channels ;
 
 	length = mpg123_length (mh) ;
+	if (length == 0)
```
arthurt, line 221: Please change to
```c
if (length == 0 && (!psf->is_pipe))
```
Cannot do a scan on a pipe.

src/mpeg_decode.c
```diff
@@ -218,6 +218,11 @@ mpeg_dec_fill_sfinfo (mpg123_handle *mh, SF_INFO *info)
...
+	{	mpg123_scan(mh);
```
arthurt, line 222: Please check for errors from this function
```c
if ((error = mpg123_scan) != MPG123_OK)
    return error
```

arthurt (changes requested):

evpobr:
Hi @JJL772 .

JJL772:
Thanks for the clarification! I've applied the suggestions and updated the commit message to be more accurate.

The files I encountered were from the game Portal 2, and seem to just be a pile of MP3 frames with no additional info:
[a2_triple_laser03.zip](https://github.com/libsndfile/libsndfile/files/11780780/a2_triple_laser03.zip)

To clarify, sndfile opens these files successfully, but won't read any samples with `sf_read` because it thinks the length is 0.

src/mpeg_decode.c
```diff
@@ -218,6 +218,12 @@ mpeg_dec_fill_sfinfo (mpg123_handle *mh, SF_INFO *info)
 	info->channels = channels ;
 
 	length = mpg123_length (mh) ;
+	if (length == 0 && !psf->is_pipe)
```
sezero, line 221: Shouldn't `length == 0` be `length <= 0` instead?

JJL772: Yes I think so, thanks!

arthurt (approved):

PR #959, opened 2023-09-05
Fix truncation for float to PCM

Sile-Yin-Bose:
Instead of floor, float to int quantization should round to nearest.

In current `pcm.c`, all floating point to fixed point clipping array are scaling the number to 32-bit integer range and converted to `int` then truncate the smaller bits. This causes the number being rounded to floor instead of nearest integer. Change this to scaling to corresponding range for target data type and truncate the higher bits would fix this.

### DIFF
--- a/src/pcm.c
+++ b/src/pcm.c
@@ -1782,20 +1782,20 @@ static void
 f2sc_clip_array (const float *src, signed char *dest, int count, int normalize)
 {	float	normfact, scaled_value ;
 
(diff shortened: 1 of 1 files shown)


arthurt:
Thanks for your contribution!

I added a two line missed change that was failing the tests, then merged your changes on a different branch.

Usually GitHub is smart enough to notice these things, and mark as merged, but in this case it isn't. So closing, but this was merged in 7662bda7a

PR #961, opened 2023-09-09
pcm: Remove CPU clipping

arthurt:
Changes introduced in #959 remove scaling to 32-bit ints before reduction to output integer width, so no CPU clipping would occur. This change ensures clipping happens for smaller than 32-bit int outputs.

In my opinion the tradeoff of losing automatic clipping that happens sometimes on not all CPUs is far outweighed by the better conversion offered by scaling to range in floating point first.

More generally though, how wise is it to maintain the CPU clipping logic? I seem to remember there was another place it was used where it did not work.


SoapGentoo:
@arthurt can we remove the autoconf macro for this as well?

arthurt:
@SoapGentoo I haven't removed all the CPU clipping uses, just the ones rendered wrong by #959. But, that is my question, should we just remove it altogether?

SoapGentoo:
I personally would, do we really care about some arb embedded processors without FPUs?

evpobr (approved):

SoapGentoo (approved):

arthurt:
Actually, looking at the clip macro, we don't detect it correctly in the common SSE2 case anyways.

The M4 macro uses lrint(), but the code uses SSE2 _mm_cvtss_si32 is available. lrint() doesn't clip generally. There is a SSE2 instruction _mm_cvt**t**ss_si32 that does clip, but it's not the one we are using.

arthurt:
Updated to remove all detection and use of cpu clipping.

evpobr:
@umlaeute, сan you take a look at this?

umlaeute:
i don't really care (as long as it still works "as expected" on *all* platforms).

@SoapGentoo wrote:
> I personally would, do we really care about some arb embedded processors without FPUs?

so i think we *should* care about these systems. "caring" doesn't mean that we need to ensure that there are optimized code paths for these platforms, but that it ("reading/writing soundfiles") actually *works*.


a casual glance at the changes looks good :-)

PR #976, opened 2023-10-09
Update ossfuzz header to fuzz all file modes

corydu:
The current OSS-Fuzz fuzzers are only fuzzing the `SFM_READ` file. This commit also adds the ability to fuzz the `SFM_WRITE` and `SFM_RDWR` file modes as well.


evpobr:
Hi @corydu

Issue #833, opened 2022-04-29
UndefinedBehaviorSanitizer: multiple signed integer overflow


pietroborrello:
##### **Describe the bug**
UndefinedBehaviorSanitizer: multiple signed integers overflow in the codebase. I attach different testcases that trigger different overflows.

##### **To Reproduce**
Built libsndfile using clang-10 according to [the oss-fuzz script](https://github.com/google/oss-fuzz/blob/master/projects/libsndfile/build.sh) with `CXXFLAGS='-O1 -fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,object-size,return,returns-nonnull-attribute,shift,signed-integer-overflow,unreachable,vla-bound,vptr'`

commit: 4b01368fd8a1e4541d7a5d125f70e99d30851718

##### **Example UBSAN Output**

```
$ ./sndfile_alt_fuzzer id:000006,sig:06,src:001126,time:22443,op:havoc,rep:4,trial:0
INFO: Seed: 3120870912
INFO: Loaded 1 modules   (33759 inline 8-bit counters): 33759 [0x8977c3, 0x89fba2), 
INFO: Loaded 1 PC tables (33759 PCs): 33759 [0x6f6b48,0x77a938), 
sndfile_alt_fuzzer: Running 1 inputs 1 time(s) each.
Running: id:000006,sig:06,src:001126,time:22443,op:havoc,rep:4,trial:0
src/au.c:324:54: runtime error: signed integer overflow: 1684960000 + 779316836 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/au.c:324:54 in 
src/au.c:326:29: runtime error: signed integer overflow: 1684960000 + 779316836 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/au.c:326:29 in 
src/au.c:327:40: runtime error: signed integer overflow: 1684960000 + 779316836 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/au.c:327:40 in 
Executed id:000006,sig:06,src:001126,time:22443,op:havoc,rep:4,trial:0 in 1 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***

or for example

[...]

Issue #789, opened 2021-11-14
Signed integer overflow in src/mat4.c:323


shao-hua-li:
Hi there,

I found an undefined behavior in src/mat4.c:323, which is a signed integer overflow.

- libsndfile version: commit c7b69d7589d70de3765d0aab15dad8d6b307d3c7
- Compile args: CFLAGS='-fsanitize=undefined' ./configure --disable-shared && make
- Compiler: clang13
- Platform: Ubuntu 20.04.2 LTS, x86_64
- Reproduce: `./programs/sndfile-metadata-get mat4.c_int_overflow`
- POC:  [mat4.c_int_overflow.tar.gz](https://github.com/libsndfile/libsndfile/files/7534888/mat4.c_int_overflow.tar.gz)

Undefined Behavior Sanitizer report:
```
src/mat4.c:323:41: runtime error: signed integer overflow: 205 * -100663296 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat4.c:323:41 in
src/mat4.c:323:48: runtime error: signed integer overflow: 838860800 * 4 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat4.c:323:48 in
```

A possible quick fix is to add explicit type casting in mat4.c:323, like `psf->dataend = psf->dataoffset + (long)rows * cols * psf->bytewidth ;`

Another consequence of this bug is for `clang -O1` and `clang -O2`, the `frame` filed in outputs would be different.

arthurt:
mat4.c
```
	int		rows, cols, imag ;
...
		psf->dataend = psf->dataoffset + rows * cols * psf->bytewidth ;
```

So clang is complaining that the result of `rows * cols` overflows a 32-bit int. `psf->dataend` is a sf_count_t, so 64-bits. A change of bracketing, or casting rows and cols before the multiplication would fix this.

stiepan:
Hi,
Are there any plans/timeline to address this issue and the [CVE-2022-33065](https://nvd.nist.gov/vuln/detail/CVE-2022-33065) vulnerability?

PR #979, opened 2023-10-17
Fix various numeric overflow vulns throughout code (CVE-2022-33065)

amstewart:
Issues #789 and #833 alert that the clang fuzzer reports several numeric overflow possibilities throughout the codebase, which can lead to undefined behaviors. MITRE opened [CVE-2022-33065](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-33065) against these issues, and has issued them a high severity. This PR attempts to fix all the outstanding numeric overflows, as detected by the ossfuzzer tool, and the reproducing case from #789.

```
src/au.c:324:54: runtime error: signed integer overflow: 1684960000 + 779316836 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/au.c:324:54 in 
src/au.c:326:29: runtime error: signed integer overflow: 1684960000 + 779316836 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/au.c:326:29 in 
src/au.c:327:40: runtime error: signed integer overflow: 1684960000 + 779316836 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/au.c:327:40 in 
--
src/mat4.c:323:41: runtime error: signed integer overflow: -587202559 * 553648128 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat4.c:323:41 in 
src/mat4.c:323:48: runtime error: signed integer overflow: 553648128 * 8 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat4.c:323:48 in 
src/mat4.c:107:35: runtime error: signed integer overflow: 8 * -587202559 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat4.c:107:35 in 
--
src/mat4.c:107:35: runtime error: signed integer overflow: 8 * -452984831 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat4.c:107:35 in 
--
src/avr.c:165:31: runtime error: signed integer overflow: 1862270847 * 2 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/avr.c:165:31 in 
--
src/sds.c:457:46: runtime error: signed integer overflow: 2113929216 + 66846720 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/sds.c:457:46 in 
--
src/sds.c:457:46: runtime error: signed integer overflow: 2113929216 + 60030976 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/sds.c:457:46 in 
--
src/sds.c:457:46: runtime error: signed integer overflow: 2113929216 + 66846720 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/sds.c:457:46 in 
--
src/nms_adpcm.c:261:20: runtime error: signed integer overflow: 27648 * 78866 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/nms_adpcm.c:261:20 in 
src/nms_adpcm.c:261:33: runtime error: signed integer overflow: -2114480128 + -943374336 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/nms_adpcm.c:261:33 in 
--
src/ircam.c:181:40: runtime error: signed integer overflow: -2147483521 * 4 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/ircam.c:181:40 in 
src/pcm.c:130:35: runtime error: signed integer overflow: 4 * -2147483521 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/pcm.c:130:35 in 
--
src/nms_adpcm.c:1094:39: runtime error: signed integer overflow: 409044504 * 160 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/nms_adpcm.c:1094:39 in 
--
src/common.c:1155:17: runtime error: signed integer overflow: 2 + 2147483646 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/common.c:1155:17 in 
--
src/ima_adpcm.c:241:44: runtime error: signed integer overflow: 64 * -34645767 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/ima_adpcm.c:241:44 in 
--
src/rf64.c:245:39: runtime error: signed integer overflow: 9223372036854775807 + 8 cannot be represented in type 'long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/rf64.c:245:39 in 
--
src/sds.c:457:46: runtime error: signed integer overflow: 2113929216 + 66846720 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/sds.c:457:46 in 
--
src/aiff.c:1709:8: runtime error: signed integer overflow: 16 + 2147483637 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/aiff.c:1709:8 in 
--
src/mat5.c:117:35: runtime error: signed integer overflow: 8 * -1294148387 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/mat5.c:117:35 in
```

I tried to limit the impact of this patchset as much as is possible. But I'm just passing through this project and have very little domain knowledge. I would appreciate an ACK from the code experts that my changes make sense. I'll mark the changes which I think are non-trivial in commentary below.

# Testing
This patchset was developed and tested primarily on clang 15. I have tested somewhat with clang 13 as well, which behaved consistently. I found Google's `gcr.io/oss-fuzz-base/base-builder` container provided the best toolchain to run the ossfuzzer.

* [x] `make check` passes with this patchset applied.
  * Log attached as `check.f.log`. The `UndefinedBehaviorSanitizer` warnings from `cue_test_var 101` reproduce with and without this patchset, and are probably expected - given that the test configuration seems to be testing a purposefully invalid case.
* [x] With this patchset in place, `ossfuzz/ossfuzz.sh` runs without reporting any of the numeric overflow warnings from above or from #833.
  * Log attached as `ossfuzzer.f.log`. The various nonfatal `Error {1,A}` lines reproduce on my machine with and without this patchset. I'm not sure what they are supposed to mean. But they seem unrelated.
* [x] Running `libsnfile-metadata-get` on the reproducing test load provided in #789 no longer produces the sanitizer warning reported in the issue.

That's all the testing that I know to run. If there are other cases that I should check, please let me know.

# Merging
This patchset includes some minor fixes to the ossfuzz tooling, that I had to make to get it working on my machine. I expect they will partially conflict with #976. I can rebase and adjust depending on which of us is pulled first.

[check.f.log](https://github.com/libsndfile/libsndfile/files/12959651/check.f.log)
[ossfuzzer.f.log](https://github.com/libsndfile/libsndfile/files/12959668/ossfuzzer.f.log)

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -18,6 +18,7 @@
 
 #include <config.h>
 
+#include <limits.h>
 #include <stdarg.h>
 #include <string.h>
 #if HAVE_UNISTD_H
@@ -990,6 +991,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 	double			*doubleptr ;
 	char			c ;
 	int				byte_count = 0, count = 0 ;
+	int				read_bytes = 0 ;
 
 	if (! format)
 		return psf_ftell (psf) ;
@@ -998,6 +1000,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 
 	while ((c = *format++))
 	{
+		read_bytes = 0 ;
 		if (psf->header.indx + 16 >= psf->header.len && psf_bump_header_allocation (psf, 16))
 			break ;
 
@@ -1014,7 +1017,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 					intptr = va_arg (argptr, unsigned int*) ;
 					*intptr = 0 ;
 					ucptr = (unsigned char*) intptr ;
-					byte_count += header_read (psf, ucptr, sizeof (int)) ;
+					read_bytes = header_read (psf, ucptr, sizeof (int)) ;
 					*intptr = GET_MARKER (ucptr) ;
 					break ;
 
@@ -1022,7 +1025,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 					intptr = va_arg (argptr, unsigned int*) ;
 					*intptr = 0 ;
 					ucptr = (unsigned char*) intptr ;
-					byte_count += header_read (psf, sixteen_bytes, sizeof (sixteen_bytes)) ;
+					read_bytes = header_read (psf, sixteen_bytes, sizeof (sixteen_bytes)) ;
 					{	int k ;
 						intdata = 0 ;
 						for (k = 0 ; k < 16 ; k++)
@@ -1034,14 +1037,14 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 			case '1' :
 					charptr = va_arg (argptr, char*) ;
 					*charptr = 0 ;
-					byte_count += header_read (psf, charptr, sizeof (char)) ;
+					read_bytes = header_read (psf, charptr, sizeof (char)) ;
 					break ;
 
 			case '2' : /* 2 byte value with the current endian-ness */
 					shortptr = va_arg (argptr, unsigned short*) ;
 					*shortptr = 0 ;
 					ucptr = (unsigned char*) shortptr ;
-					byte_count += header_read (psf, ucptr, sizeof (short)) ;
+					read_bytes = header_read (psf, ucptr, sizeof (short)) ;
 					if (psf->rwf_endian == SF_ENDIAN_BIG)
 						*shortptr = GET_BE_SHORT (ucptr) ;
 					else
@@ -1051,7 +1054,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 			case '3' : /* 3 byte value with the current endian-ness */
 					intptr = va_arg (argptr, unsigned int*) ;
 					*intptr = 0 ;
-					byte_count += header_read (psf, sixteen_bytes, 3) ;
+					read_bytes = header_read (psf, sixteen_bytes, 3) ;
 					if (psf->rwf_endian == SF_ENDIAN_BIG)
 						*intptr = GET_BE_3BYTE (sixteen_bytes) ;
 					else
@@ -1062,7 +1065,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 					intptr = va_arg (argptr, unsigned int*) ;
 					*intptr = 0 ;
 					ucptr = (unsigned char*) intptr ;
-					byte_count += header_read (psf, ucptr, sizeof (int)) ;
+					read_bytes = header_read (psf, ucptr, sizeof (int)) ;
 					if (psf->rwf_endian == SF_ENDIAN_BIG)
 						*intptr = psf_get_be32 (ucptr, 0) ;
 					else
@@ -1072,7 +1075,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 			case '8' : /* 8 byte value with the current endian-ness */
 					countptr = va_arg (argptr, sf_count_t *) ;
 					*countptr = 0 ;
-					byte_count += header_read (psf, sixteen_bytes, 8) ;
+					read_bytes = header_read (psf, sixteen_bytes, 8) ;
 					if (psf->rwf_endian == SF_ENDIAN_BIG)
 						countdata = psf_get_be64 (sixteen_bytes, 0) ;
 					else
@@ -1083,7 +1086,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 			case 'f' : /* Float conversion */
 					floatptr = va_arg (argptr, float *) ;
 					*floatptr = 0.0 ;
-					byte_count += header_read (psf, floatptr, sizeof (float)) ;
+					read_bytes = header_read (psf, floatptr, sizeof (float)) ;
 					if (psf->rwf_endian == SF_ENDIAN_BIG)
 						*floatptr = float32_be_read ((unsigned char*) floatptr) ;
 					else
@@ -1093,7 +1096,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 			case 'd' : /* double conversion */
 					doubleptr = va_arg (argptr, double *) ;
 					*doubleptr = 0.0 ;
-					byte_count += header_read (psf, doubleptr, sizeof (double)) ;
+					read_bytes = header_read (psf, doubleptr, sizeof (double)) ;
 					if (psf->rwf_endian == SF_ENDIAN_BIG)
 						*doubleptr = double64_be_read ((unsigned char*) doubleptr) ;
 					else
@@ -1117,7 +1120,7 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 					charptr = va_arg (argptr, char*) ;
 					count = va_arg (argptr, size_t) ;
 					memset (charptr, 0, count) ;
-					byte_count += header_read (psf, charptr, count) ;
+					read_bytes = header_read (psf, charptr, count) ;
 					break ;
 
 			case 'G' :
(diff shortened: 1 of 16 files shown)


src/sds.c
```diff
@@ -454,7 +454,7 @@ sds_2byte_read (SF_PRIVATE *psf, SDS_PRIVATE *psds)
 
 	ucptr = psds->read_data + 5 ;
 	for (k = 0 ; k < 120 ; k += 2)
-	{	sample = arith_shift_left (ucptr [k], 25) + arith_shift_left (ucptr [k + 1], 18) ;
+	{	sample = arith_shift_left (ucptr [k], 25) | arith_shift_left (ucptr [k + 1], 18) ;
```
amstewart, line 457: @ codeowners
Based on my reading of what the sampling loop is trying to do. I *think* a bitwise OR is really what should have been done in the first place. Arithmetically, these should be equivalent, **unless** the ucptr bytes are supposed to be interpreted as signed values. I don't know why they would be, but I'll stand to be wrong.

src/common.h
```diff
@@ -439,7 +439,7 @@ typedef struct sf_private_tag
 	sf_count_t		datalength ;	/* Length in bytes of the audio data. */
 	sf_count_t		dataend ;		/* Offset to file tailer. */
 
-	int				blockwidth ;	/* Size in bytes of one set of interleaved samples. */
+	sf_count_t		blockwidth ;	/* Size in bytes of one set of interleaved samples. */
```
amstewart, line 442: @ codeowners
It seemed sensible to change the blockwidth attribute to be an `sf_count_t`, since most places that it is used throughout the codebase assign it to be the product of int32s. But this change does substantially alter the dimensionality of this struct. Does it seem like this is a breaking API change? Should we avoid doing this?

src/common.c
```diff
@@ -1164,8 +1167,17 @@ psf_binheader_readf (SF_PRIVATE *psf, char const *format, ...)
 				psf->error = SFE_INTERNAL ;
 				break ;
 			} ;
+
+		if (read_bytes > 0 && byte_count > (INT_MAX - read_bytes))
+		{	psf_log_printf (psf, "Header size exceeds INT_MAX. Aborting.", c) ;
+			psf->error = SFE_INTERNAL ;
+			break ;
+		} else
+		{	byte_count += read_bytes ;
```
amstewart, line 1176: @ codeowners
As outlined in the commit summary, this change adds a new error case for this function. But that is probably preferable to the undefined behavior that existed previously.

src/nms_adpcm.c
```diff
@@ -258,7 +260,12 @@ nms_adpcm_update (struct nms_adpcm_state *s)
 		} ;
 
 	/* Compute the signal estimate. */
-	s->s_e = s->a [0] * s->s_r [0] + s->a [1] * s->s_r [1] + s->s_ez ;
+	signal_estimate = (int64_t) s->a [0] * s->s_r [0] + (int64_t) s->a [1] * s->s_r [1] + s->s_ez ;
+	if (signal_estimate > INT_MAX)
+		signal_estimate = INT_MAX ;
+	else if (signal_estimate < -INT_MAX)
+		signal_estimate = -INT_MAX ;
+	s->s_e = (int) signal_estimate ;
```
amstewart, line 268: @ codeowners
I'm making a bold assertion that it is logically correct to clamp the signal estimate value here, rather than erroring. Let me know if that seems incorrect.

amstewart:
I also see a couple places in the commit summaries where I erroneously wrote `st_count_t`, instead of `sf_count_t`. Will fix in a patch V2.

evpobr:
Hi @amstewart . @arthurt , @SoapGentoo , can you take a look?

src/nms_adpcm.c
arthurt: I'm going to push back on this one a little. I think this is a false-positive.

The signal estimate going out of bounds indicates that bad input data, so the result that comes out being undefined is ok. In essence, any fix here would be "you put garbage in, what kind of garbage out would you like?"

Critical to this is that NMS ADPCM has no variable control flow: there are no branches. The same functions are called always in the same order, all loops are constant (could be unrolled), and all array access is bounded. Static analysis could inline all functions and turn `nms_adpcm_decode_sample()` and `nms_adpcm_encode_sample()` each into one non-branching function. That they are functions is because the same code is used but in reverse order for encode and decode.

Is there a way to tell the fuzzer to ignore this? Or a choice of datatypes that makes this more explicit?

The choice of `int` as a datatype for the state variables was one of efficiency. "If you don't know what to pick, pick `int`". The only requirement is that they be signed and at least 16 bit. `int_fast16_t` would be a more explicit choice.

The signal estimate and signal zero estimate calculation is actually 16-bit decimal multiplication in fixed-point 32 bits.

src/common.h
evpobr: I guess no, it is private struct, we use `SNDFILE*` as opaque pointer:

https://github.com/libsndfile/libsndfile/blob/7ed4773251c26653b5ee1a0815bfe8b213c14b54/include/sndfile.h#L357-L359

src/sds.c
amstewart: >Can you include identical changes to sds_3byte_read() and sds_4byte_read() please.

Will do.

amstewart:
# Patch V2
* [x] Fixed a few spelling mistakes in some commit summaries. (`s/st_count_t/sf_count_t`)
* [x] Expanded the bitwise-OR implementation in `sds.c:sds_2byte_read()` to the 3byte and 4byte flavors of that function.
* [x] Replaced the signal clamping implementation in the nms_adpcm module with an approach that simply pre-casts the component signals to `fast16` types.
* [x] Revalidated that `make check` and the ossfuzzer still pass as before.

src/sds.c
amstewart: Fixed patch V2.

src/nms_adpcm.c
amstewart: That's fair. If you would really prefer to ignore the error, I can investigate if that is possible. Patch V2 just cast the signal components to satisfy the fuzzer. Let me know what you want to do.

amstewart:
https://github.com/libsndfile/libsndfile/actions/runs/6578627832/job/17873683230?pr=979

PR build shows a couple test failures on macos, that seem localized to the nms_adpcm implementation. Will investigate, but I don't have a mac available so I'll have to test in the pipeline.

src/nms_adpcm.c
arthurt: Have a look at cb65f952ad1d5be7d4ba41d6c3f1c3ff9e0916b2 :)

amstewart: Yep. That looks good to me. I'll replace my implementation with that in V3 (if only to keep the CVE commits relatively clean.) Thanks for the expertise!

amstewart:
# Patch V3
* [x] Incorporated @arthurt's patch to reduce the general size of the `nms_adpcm` state variables. Which I also expect will fix the macos CI tests. (https://github.com/libsndfile/libsndfile/pull/979/commits/6e162cb767e81cd15f4dc2a2fa253d2e36adfd70)

arthurt (approved):

evpobr:
@SoapGentoo ?

SoapGentoo (approved):

evpobr (approved):

evpobr:
Thank you all!

PR #988, opened 2023-12-21
Fix sign of ulaw decompression

LWinterberg:
7f apparently should resolve to -1 and not 0. That at least matches the table I see in https://web.archive.org/web/20110719132013/http://hazelware.luggle.com/tutorials/mulawcompression.html 

It probably doesn't matter for most intents and purposes, but we did have someone run into it: https://github.com/audacity/audacity/issues/5788


evpobr:
Hi @LWinterberg.

LWinterberg:
Ah, I'm afraid I'm not entirely understanding how your tests function. I'm not technically a programmer anyway :) 

I did find the relevant table in the specification in the meantime though: 

image

It seems like the *decision value* is -1, but the *output value* is 0. So maybe the source I initially was following (at the bottom of the µlaw wikipedia article) got it wrong?

Issue #768, opened 2021-07-11
Ability To Set Loop Info


arthurt:
Received an enquiry from a user about the ability to set loop info in a file. This is not currently possible. Support for this would have to be per-major-format.

Major formats that currently return loop info are:
- WAV
- AIFF

Major format that has loop info that isn't implemented:
- XI

There may be complications related to ACID chunks. The mapping from SF_LOOP_INFO to the information in an ACID chunk is not quite 1:1.

PR #997, opened 2024-02-20
Feature/#768 sfc set loop info

riban-bw:
Adds ability to write loop info using command SFC_SET_LOOP_INFO. This resolves issue #768.

I have checked various scenarios and this does seem to work for the file formats that support loop info and does not seem to upset the other file formats.


evpobr:
Hi @riban-bw . @arthurt , @SoapGentoo , need your advice.

PR #1019, opened 2024-06-05
flac: Add support for 32-bit FLAC files

robertbcalhoun:
* Support reading and writing of 32-bit PCM audio in FLAC format.
* Add read/write test for 32-bit FLAC, and clip tests for all FLAC formats.
* Bump the versions of flac and opus used in statically-linked libraries.
* Update support matrix in formats.md.

Fixes: https://github.com/libsndfile/libsndfile/discussions/910
Fixes: https://github.com/libsndfile/libsndfile/discussions/980


evpobr:
Hi @robertbcalhoun .

evpobr:
Hmm, some tests failed.

robertbcalhoun:
I tested on Debian bookworm x86-64 and Windows 10 x86-64 (using vcpkg); all tests pass on both. I did not test on MacOS or any ARM cpu. 

The build is failing tests on the CI server because the version of Ubuntu being used for builds has flac version 1.3.3, and support for 32-bit flac files requires flac version 1.4.0 or better. It is a bug in the PR that the requirement of flac > 1.4.0 wasn't flagged at the configuration step.

I could use some help identifying where the required version should be specified. Here's what I did (and did not) do in the PR:

* I updated required flac version when building static libs in [static-deps-build.mk](https://github.com/libsndfile/libsndfile/pull/1019/commits/d6bca67038538081beb6a4c366744c2bb3b37206#diff-9222ad3e45d75733d8125020517442ea040d69b05f2dcc1ba15cac573a691e60).
* I updated the required flac version when using autotools in [configure.ac](https://github.com/libsndfile/libsndfile/pull/1019/commits/d6bca67038538081beb6a4c366744c2bb3b37206#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810).
* The [vcpkg manifest for libflac](https://vcpkg.io/en/package/libflac.html) will build 1.4.3, so I don't think anything needs to be done there.
* I missed adding the requirement to [cmake/FindFLAC.cmake](https://github.com/libsndfile/libsndfile/blob/master/cmake/FindFLAC.cmake), which currently does not specify a required package version. If pkgconfig is used, we can add the requirement with diff below change. I don't know how to add the requirement to CMake if pkgconfig is not in use.
```diff --git a/cmake/FindFLAC.cmake b/cmake/FindFLAC.cmake
index 3d415d76..6b1c87a3 100644
--- a/cmake/FindFLAC.cmake
+++ b/cmake/FindFLAC.cmake
@@ -14,7 +14,7 @@ endif ()
 find_package (Ogg QUIET)
 
 find_package (PkgConfig QUIET)
-pkg_check_modules(PC_FLAC QUIET flac)
+pkg_check_modules(PC_FLAC QUIET flac>1.4.0)
```
Any other places dependency versions need to be specified?

Once the configuration issue are resolved, there remains the philosophical question for the maintainers on what versions of dependencies might reasonably be demanded. The current LTS release of Ubuntu (noble, 24.04LTS) ships with flac 1.4.3, as does the previous non-LTS version of Ubuntu (mantic, 23.10) and the "current stable" version of Debian (bookworm). However, the previous Ubuntu LTS release (jammy, 22.04LTS) and the "old stable" version of Debian (bullseye) ship with flac 1.3.3, which won't work.

In my opinion, it is reasonable to expect anyone building or installing the latest version of libsndfile have the latest (stable)  versions of the required dependencies, but this is up to the maintainers.

robertbcalhoun:
* Added libflac version checks to cmake builds.
* Changed github actions testrunner from `ubuntu-latest` (currently 22.04) to `ubuntu-24.04` Ubuntu 24.04, which is [now available in beta.](https://github.blog/changelog/2024-05-14-github-hosted-runners-public-beta-of-ubuntu-24-04-is-now-available/).

Ubuntu 24.04 is an appropriate CI environment for the newest versions of libsndfile, since this is the Ubuntu branch most likely to accept new versions of libsndfile. Ubuntu 24.04 already ships with libflac 1.4.3, which means it is possible to create a 32-bit FLAC file with `flac.exe` under Ubuntu 24.04 that `libsndfile` under Ubuntu 24.04 cannot read.

PR #1027 must be merged before libsndfile will build correctly under ubuntu-24.04.

evpobr (changes requested):
@robertbcalhoun , i guess we need to enable 32-bit FLAC support only if version >=1.4.0 is detected, otherwise just build a library without 32-bit support. If the distribution has an old version of FLAC, this is certainly not our problem, but libsndfile is designed to work without supporting some of the functionality, especially if it depends on external libraries.

robertbcalhoun:
OK, I will modify these commits to support various versions of libFLAC.

Could you consider incorporating at least https://github.com/libsndfile/libsndfile/pull/1027/commits/d2ba05a19f9038f9c5366af8c831f24cbe573a20 of #1027? At present it is not possible to compile libsndfile under Ubuntu 24.04 without gcc throwing warnings, and the warnings are justified because the test data is being corrupted in the current code. (The test passes because metadata read ==  metadata write, but the data being written are not what was intended.)

jesperpedersen:
@robertbcalhoun Hi, did you have a chance to look at this ?

Meowchestra:
Also would love to see SF_FORMAT_PCM_32 FLAC supported in libsndfile since libflac has it. Hopefully won't take too many more years. 😢

robertbcalhoun:
Hi libsndfile maintainers,

This PR has been refactored against current codebase. The number of changes is smaller and it should now work properly on all platforms.

The recent (2026-08-19) CI/CD checks failed because of Github Actions problems rather than commit content. I believe the CI/CD issues were fixed in PR #1159 . This PR is ready to have CI/CD tests re-run and code reviews.

Per comments above, there is user interest in this feature and it would be good to incorporate it in libsndfile. I've been using FLAC32 with my own version of libsndfile for two years and it has been working fine.

Thank you!

nigels-com:
I'm happy to do a detailed review, and throw some Claude tokens at it also, if it gives the maintainers some additional confidence about merging this.

Issue #1012, opened 2024-05-23
Heap-buffer-overflow in sndfile.c


xiaobaozidi:
Dear Maintainer,

We have identified a Heap-buffer-overflow memory bug in sndfile.c in commit [c81375f070f3c676496](https://github.com/libsndfile/libsndfile/commit/c81375f070f3c6764969a738eacded64f53a076e)

To reproduce the bug:

compile with asan:
```
export CFLAGS="-fsanitize=address -fno-omit-frame-pointer -g"
export CXXFLAGS="-fsanitize=address -fno-omit-frame-pointer" 
```
and run:

`./sndfile-convert POC out.wav`
[poc_overflow.zip](https://github.com/libsndfile/libsndfile/files/15424786/poc_overflow.zip)

Please find the detailed report below:

==3760181==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000020 at pc 0x000000498c97 bp 0x7fffc34ef030 sp 0x7fffc34ee7f8
READ of size 64 at 0x602000000020 thread T0
    #0 0x498c96 in __asan_memcpy (/benchmark/libsndfile/out/afl/sndfile-convert+0x498c96)
    #1 0x4ed8bb in sf_command /benchmark/libsndfile/src/sndfile.c
    #2 0x4cc061 in copy_metadata /benchmark/libsndfile/programs/sndfile-convert.c:396:7
    #3 0x4cc061 in main /benchmark/libsndfile/programs/sndfile-convert.c:351:2
    #4 0x7fa3c1f2ac86 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x21c86)
    #5 0x41fb69 in _start (/benchmark/libsndfile/out/afl/sndfile-convert+0x41fb69)

0x602000000020 is located 0 bytes to the right of 16-byte region [0x602000000010,0x602000000020)
allocated by thread T0 here:
    #0 0x49983d in malloc (/benchmark/libsndfile/out/afl/sndfile-convert+0x49983d)
    #1 0x5130cf in aiff_read_chanmap /benchmark/libsndfile/src/aiff.c:1797:27
    #2 0x5130cf in aiff_read_header /benchmark/libsndfile/src/aiff.c:885:19
    #3 0x5130cf in aiff_open /benchmark/libsndfile/src/aiff.c:249:17
    #4 0x4da9f6 in psf_open_file /benchmark/libsndfile/src/sndfile.c:3169:13
[...]

PR #1032, opened 2024-07-18
Fix Heap-buffer-overflow in sndfile.c #issue1012

haruki3hhh:
Try to fix issue: https://github.com/libsndfile/libsndfile/issues/1012

### DIFF
--- a/src/aiff.c
+++ b/src/aiff.c
@@ -1794,7 +1794,7 @@ aiff_read_chanmap (SF_PRIVATE * psf, unsigned dword)
 
 		free (psf->channel_map) ;
 
-		if ((psf->channel_map = malloc (chanmap_size)) == NULL)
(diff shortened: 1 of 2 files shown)


src/aiff.c
arthurt, line 1: Could you share your reasoning for changing the malloc argument at line 1797, and not `chanmap_size` at line 1793?

This change means the allocated memory and `chanmap_size` could differ.

src/sndfile.c
```diff
@@ -1421,13 +1421,19 @@ sf_command	(SNDFILE *sndfile, int command, void *data, int datasize)
 
 		case SFC_GET_CHANNEL_MAP_INFO :
 			if (psf->channel_map == NULL)
-				return SF_FALSE ;
+				return SF_FALSE;
```
arthurt, line 1424: Formatting error, please fix. This project leaves a space before the final semicolon.

arthurt (changes requested):

PR #1048, opened 2024-11-27
Unseekable Steams Support (pipes) for Virtual IOs

mmalygin:
If your stream is not seekable, set "seek", "get_filelen", and "tell" callbacks to NULL in SF_VIRTUAL_IO. In this case, the library will treat your stream as a non-seekable virtual pipe. Note that not all operations are available for pipes. For example you can open a WAV file for read but it's impossible to write WAV file to a pipe because the WAV format requires a length of file to be written at beginning of the file which is impossible without supporting of "seek".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+* Unseekable steams support (pipes) for Virtual IOs
+
+  If your stream is not seekable, set "seek", "get_filelen", and "tell" callbacks
+  to NULL in SF_VIRTUAL_IO. In this case, the library will treat your stream as
+  a non-seekable virtual pipe. Note that not all operations are available for pipes.
+  For example you can open a WAV file for read but it's impossible to write WAV file
+  to a pipe because the WAV format requires a length of file to be written at beginning
+  of the file which is impossible without supporting of "seek".
+
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


arthurt:
Good idea. I just want to do some more testing before giving the okay. What do you think @erikd, @evpobr, @SoapGentoo?

As a tangent, what are people's thoughts about the virtual IO read and write functions? Currently the read and write functions are basically a user-specified replacement to `psf_fread()` and `psf_fwrite()` rather than for `read()` and `write()` on an FD. This means IO retry, read size and interrupt behaviour have to be handled by the VIO implementations.

This does increase the flexibility for implementers, but it also means that the behaviour of `psf_fread()` and `psf_fwrite()` can be changed substantially by VIO. Consider that non-VIO `psf_fread()` never returns < 0, but a VIO implementation can. This breaks all kinds of error handling in the rest of the sndfile code.

mmalygin:
@arthurt Below is just my 50 cents.
The issue here is not in VIO operations but in `psf_read()/psf_write()/psf_seek()` and `psf_tell()` themselves. They do not have error handling for IO operations. What will be if, for example, `read(..)` in non-VIO `psf_read()` returns -1 and `errno` is `EIO`? This can happen, for example, on a network file system like NFS.
These functions must return something like -1 for IO errors and all code that uses these functions must check for their return value and if it's -1, break further processing with an error as well.
The same is valid for VIO. Any IO operation can produce errors.

mmalygin:
@arthurt @erikd @SoapGentoo @evpobr
Guys, do you have any comments/suggestions? Can we merge this PR?

Issue #1054, opened 2024-12-17
build shared library and enable tests [WIP PATCH]


LocutusOfBorg:
Hello, I found a big regression with cmake w.r.t. autotools, and I crafted a patch to build the testsuite and rely on static library always, not exposed.

Please have a look at this patch, I would like to know your thoughts before cleaning it up

```
Description: <short summary of the patch>
Author: Gianfranco Costamagna <locutusofborg@debian.org>
Origin: (upstream|backport|vendor|other), (<patch-url>|commit:<commit-id>)
Bug: <upstream-bugtracker-url>
Bug-Debian: https://bugs.debian.org/<bugnumber>
Bug-Ubuntu: https://launchpad.net/bugs/<bugnumber>
Forwarded: (no|not-needed|<patch-forwarded-url>)
Applied-Upstream: <version>, (<commit-url>|commit:<commid-id>)
Reviewed-By: <name and email of someone who approved/reviewed the patch>
Last-Update: 2024-12-12

Index: libsndfile-1.2.2/CMakeLists.txt
===================================================================
--- libsndfile-1.2.2.orig/CMakeLists.txt
+++ libsndfile-1.2.2/CMakeLists.txt
@@ -68,7 +68,7 @@
 
 option (BUILD_SHARED_LIBS "Build shared libraries" OFF)
 if (BUILD_SHARED_LIBS AND BUILD_TESTING)
-	set (BUILD_TESTING OFF)
+	#set (BUILD_TESTING OFF)
 	message ("Build testing required static libraries. To prevent build errors BUILD_TESTING disabled.")
 endif ()
 option (BUILD_PROGRAMS "Build programs" ON)
@@ -256,7 +256,7 @@
 # libsndfile static library
 #
 
-add_library (sndfile
+set(SNDFILE_SOURCES
 	src/sfconfig.h
 	src/sfendian.h
 	src/sf_unistd.h
@@ -377,6 +377,7 @@
 	${CMAKE_CURRENT_BINARY_DIR}/src/config.h
 	)
 
+add_library (sndfile ${SNDFILE_SOURCES})
 add_library (SndFile::sndfile ALIAS sndfile)
 
 target_include_directories (sndfile
@@ -941,14 +942,26 @@
 	lsf_autogen (src test_endswap c)
 
 	# utils static library
-	add_library(test_utils STATIC tests/utils.c)
+	add_library(test_utils STATIC tests/utils.c ${SNDFILE_SOURCES})
[...]

PR #1058, opened 2025-01-02
cmake: Unconditional unit test support

derselbst:
Previously, compiling the unit tests was only supported when libsndfile was compiled as static library. This however prevented several distros (like Ubuntu) from migrating to libsndfile's CMake based buildsystem, as they are compiling shared libs and they still want to be able to execute the unit tests.

This PR adds an additional object library `sndfile-OBJ` to the cmake buildsystem. The unit tests link against that object lib, causing the test binaries to effectively link against all the individual object files, rather than the final static or shared lib. The previously disabled `win32_ordinal_test` has also been re-enabled. 

In addition to that I have 
* remove duplicated test code from `src/ima_oki_adpcm.c`,
* refactored `src/nms_adpcm.c` a bit, i.e. moving declaration to a new header file and making functions used by the unit test non-static, and
* unconditionally enabled compiling the unit tests in CI builds.

I successfully compiled and executed the unit tests locally for both, static and shared builds.


derselbst:
If accepted this PR would supersede #1054. @FtZPetruska @LocutusOfBorg @pedrolcl Pinging you guys, since you might be interested in this change as well.

PR #1073, opened 2025-03-28
fix cmake min...max syntax

### LINKED ISSUE #1072: Correct syntax for range in cmake is 3 dots
https://github.com/libsndfile/libsndfile/blob/17a19ab264fcf238f46104cea9af9a0a5ca3786d/CMakeLists.txt#L1

ref: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

heitbaum:
ref: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

fixes #1072


evpobr:
Thanks @heitbaum !

Issue #1076, opened 2025-05-09
build breaks with current version of cmake, version 4.0.x


sidney:
There are two problems preventing libsndfile from building when using the current version of cmake, 4.0.x

cmake version 4.0, released March 28 removes compatibility with  cmake versions older than 3.5 by requiring that the range specified by `cmake_minimum_required` have a minimum value of at least 3.5. It also requires that the max value in the range has to be at least 4.0.

The current line in `CMakeLists.txt` is

`cmake_minimum_required (VERSION 3.12...3.26)`

Changing 3.12...3.26 to be 3.5...4.0 should allow it to build under cmake 4.0, but I see that the minimum was changed from 3.5 to 3.12 on purpose for issue #1069 (and the issue for a typo that left, issue #1072)

I don't understand the reasoning behind issue #1069 why the use of Ogg_ROOT should require a minimum version number of 3.12 instead of 3.5, so I leave that question to someone who does understand it.

However, even after changing the line to
`cmake_minimum_required (VERSION 3.12...3.26)`
it still does not build with the first error resulting from use of a feature no longer supported in cmake 4.0
```
CMake Error at cmake/SndFileChecks.cmake:255 (find_package):
  By not providing "FindPythonInterp.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "PythonInterp", but CMake did not find one.

```
See an example of this error in https://github.com/TeXworks/texworks/issues/1039

I was able to build cmake with no changes to `CMakeLists.txt` by changing the build script to call cmake with the argument
`-DCMAKE_POLICY_VERSION_MINIMUM=3.5`

I don't have a PR since this change is in the build process. I think someone would have to go through fixing whatever incompatibilities with cmake 4.0 exist, with the above workaround being documented for building libsndfile until then.

sidney:
I looked at what failed when trying to build using cmake 4.0 and it was simple enough that I went ahead and made a PR after all.

PR #1077, opened 2025-05-11
Update build for using cmake 4, fixing issue #1076

sidney:
This allows building libsndfile using cmake 4.0. See issue #1076
Changes cmake_minimum_required minimum to 3.5, since  cmake 4.0 will not accept earlier versions
Changes cmake_minimum_required maximum to 4.0
Changes use of obsoleted package PythonInterp to Python which is current and is backwards compatible with older versions of cmake

Tested by building libsndfile via homebrew on macOS 13.7.5


evpobr (approved):

evpobr:
Hi @sidney , thanks for PR.

PR #1084, opened 2025-06-15
fix opus inclusion

neheb:
The pkgconfig file for Opus includes the opus subdirectory, not the raw include directory.

### DIFF
--- a/src/ogg_opus.c
+++ b/src/ogg_opus.c
@@ -162,8 +162,8 @@
 #if HAVE_EXTERNAL_XIPH_LIBS
 
 #include <ogg/ogg.h>
-#include <opus/opus.h>
-#include <opus/opus_multistream.h>
+#include <opus.h>
+#include <opus_multistream.h>
 
 #include "ogg.h"
 #include "ogg_vcomment.h"


neheb:
now that's weird.

So the pkgconfig file on Ubuntu looks like

```
# Opus codec reference implementation pkg-config file

prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: Opus
Description: Opus IETF audio codec (floating-point build)
URL: https://opus-codec.org/
Version: 1.3.1
Requires:
Conflicts:
Libs: -L${libdir} -lopus
Libs.private: -lm
Cflags: -I${includedir}/opus
```

while the cmake build does not do this?

nigels-com:
Yeah, that does seem strange.
It's been that way upstream for [9 years now](https://github.com/xiph/opus/blob/main/opus.pc.in).
But Ubuntu says otherwise:

```
$ ls -la /usr/include/opus/
total 200
drwxr-xr-x   2 root root  4096 Nov 26 19:58 .
drwxr-xr-x 116 root root 20480 Nov 26 20:00 ..
-rw-r--r--   1 root root 14847 Apr  9  2024 opus_custom.h
-rw-r--r--   1 root root 35369 Apr  9  2024 opus_defines.h
-rw-r--r--   1 root root 48453 Apr  9  2024 opus.h
-rw-r--r--   1 root root 33541 Apr  9  2024 opus_multistream.h
-rw-r--r--   1 root root 28241 Apr  9  2024 opus_projection.h
-rw-r--r--   1 root root  5157 Apr  9  2024 opus_types.h
```

```
$ cat /etc/issue
Ubuntu 24.04.3 LTS \n \l

$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/opus.pc
# Opus codec reference implementation pkg-config file

prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: Opus
Description: Opus IETF audio codec (floating-point, custom modes build)
URL: https://opus-codec.org/
Version: 1.4
Requires:
Conflicts:
Libs: -L${libdir} -lopus
Libs.private: -lm
Cflags: -I${includedir}/opus
```

PR #1085, opened 2025-06-16
Prevent linker warning about duplicate libraries

RSAsterix:
If vorbisenc is included by sndfile, it already provides libm and ogg (which causes a linker warning)


nigels-com:
Can the failing test coverage be re-tried?

Issue #1082, opened 2025-06-04
libsndfile IRCAM File Processing Buffer Overflow


20urc3:
I've discovered a crash in libsndfile's IRCAM file format parser while fuzzing with AFL++. The issue manifests as an illegal instruction (SIGILL) when processing a specially crafted file.
Bug details

When processing the attached malformed IRCAM file, libsndfile crashes with an illegal instruction at src/ircam.c:164 in the ircam_read_header function when attempting to set the sample rate.
Steps to reproduce

- Compile libsndfile with ASAN instrumentation
- Run the attached PoC file through the library
- Observe the SIGILL crash

## 

[poc_file.txt](https://github.com/user-attachments/files/20587225/poc_file.txt)

Crash information

GDB backtrace:
```
Program received signal SIGILL, Illegal instruction.
0x00005555556e53f3 in ircam_read_header (psf=0x525000000100) at src/ircam.c:164
164            psf->sf.samplerate = (int) samplerate ;
(gdb) bt
#0  0x00005555556e53f3 in ircam_read_header (psf=0x525000000100) at src/ircam.c:164
#1  ircam_open (psf=0x525000000100) at src/ircam.c:84
#2  0x000055555568872c in psf_open_file (psf=0x525000000100, sfinfo=<optimized out>) at src/sndfile.c:3202
#3  0x0000555555684bc3 in main (argc=<optimized out>, argv=<optimized out>) at harness.c:82

Disassembly around crash point:
0x5555556e53df <ircam_open+5639>:    cmp    $0x4800252d,%eax
0x5555556e53e4 <ircam_open+5644>:    mov    (%rcx),%ecx
0x5555556e53e6 <ircam_open+5646>:    movzbl (%rcx,%rax,1),%edx
0x5555556e53ea <ircam_open+5650>:    add    $0x1,%dl
0x5555556e53ed <ircam_open+5653>:    adc    $0x0,%dl
0x5555556e53f0 <ircam_open+5656>:    mov    %dl,(%rcx,%rax,1)
=> 0x5555556e53f3 <ircam_open+5659>:    ud1    0x5(%eax),%eax
0x5555556e53f8 <ircam_open+5664>:    lea    0x252d21(%rip),%rax
0x5555556e53ff <ircam_open+5671>:    mov    (%rax),%rdi
0x5555556e5402 <ircam_open+5674>:    lea    0x258833(%rip),%rax
```

## Analysis

[...]

PR #1099, opened 2025-09-03
ircam: validate samplerate/channels; bound blockwidth (CVE-2025-52194)

zaydr-al:
This hardens IRCAM header parsing to prevent the crash reported in #1082 (CVE-2025-52194).

**Key changes:**
- Validate samplerate is finite, >0, <= ``INT_MAX`` and preserve legacy (int) cast.
- Enforce final 1..``SF_MAX_CHANNELS`` after endianness resolution.
- Compute blockwidth in sf_count_t with pre-multiply guard; verify <= ``INT_MAX`` before assigning to ``SF_PRIVATE::blockwidth (int)``.
- Guard data region and frames math; avoid division by zero.

**Behavioral impact:**
- Malformed IRCAM files are rejected with SFE_* errors (no crash).

**Testing:**
- Reproduced from #1082: unpatched crashes; patched rejects cleanly.
*(We can contribute a minimal FE flag probe if the project wants a regression test.)*

Fixes: #1082


LionNatsu:
New added precondition checks look sound.

> Testing:
> Reproduced from https://github.com/libsndfile/libsndfile/issues/1082: unpatched crashes; patched rejects cleanly.
(We can contribute a minimal FE flag probe if the project wants a regression test.)

But can you really reproduce the crash? Do you have any backtrace or core dump that can confirm this?

src/ircam.c
```diff
@@ -217,13 +225,39 @@ ircam_read_header	(SF_PRIVATE *psf)
...
+	{
+		/* Pre-multiply guard ensuring bw <= SF_COUNT_MAX. */
+		if ((sf_count_t) psf->sf.channels >
+				(SF_COUNT_MAX / (sf_count_t) psf->bytewidth))
+			return SFE_MALFORMED_FILE ;
+
+		sf_count_t bw = (sf_count_t) psf->sf.channels * (sf_count_t) psf->bytewidth ;
+		if (bw <= 0 || bw > INT_MAX)
+			return SFE_MALFORMED_FILE ;
+		psf->blockwidth = (int) bw ;
```
LionNatsu, line 240: Wait. You are casting bw narrower than the original. And therefore, the range check is also wrong.

LionNatsu (changes requested):
"blockwidth" is an `sf_count_t` (defined in src/common.h), effectively `int64_t`. You should use `(sf_count_t)` and `SF_COUNT_MAX`.

src/ircam.c
zaydr-al: Updated the range check and added a pre-multiply guard with ``SF_COUNT_MAX``

src/ircam.c
```diff
@@ -159,9 +161,20 @@ ircam_read_header	(SF_PRIVATE *psf)
...
+	if (psf->sf.channels < 1 || psf->sf.channels > SF_MAX_CHANNELS)
+		return SFE_IRCAM_BAD_CHANNELS ;
+
 	psf_log_printf (psf, "marker: 0x%X\n", marker) ;
 
-	psf->sf.samplerate = (int) samplerate ;
+	/* CVE-2025-52194: validate samplerate, preserve cast-to-int semantics. */
+	if (!isfinite (samplerate) || samplerate <= 0.0f || samplerate > (float) INT_MAX)
+		return SFE_MALFORMED_FILE ;
+	{       int sr = (int) samplerate ;
```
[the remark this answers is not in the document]
zaydr-al: Fixed

LionNatsu:
LGTM, except it does not match what the CVE describes. The CVE says there should be a *memory corruption*, which is much more severe, potentially triggered by a buffer overflow *before* this type casting. So it is hardly considered a fix to the CVE.

zaydr-al:
I re-checked with sanitizers: on **unpatched**, UBSan hits the float->int cast, but **ASan-only shows no OOB/corruption** with the #1082 POC. 
With this patch the file is rejected early and sanitizers stay quiet.
I've scoped this as **hardening/mitigation**. If there's a PoC that shows a pre-cast overflow, I'm happy to test and extend.

PR #1124, opened 2026-05-18
ima_adpcm: fix int overflow when calculating sf.frames

wtay:
When calculating sf.frames, pre-cast samplesperblock to sf_count_t, to provide the calculation with enough numeric space to avoid overflows.

CVE: CVE-2026-37555


wtay:
oops duplicate

PR #1129, opened 2026-06-14
Allow individual selection of external codecs 

commodo:
Allow individual selection of external codecs
Replace the monolithic HAVE_EXTERNAL_XIPH_LIBS with per-codec HAVE_OGG,
HAVE_VORBIS, HAVE_FLAC, HAVE_OPUS, HAVE_SPEEX defines and matching
ENABLE_* CMake options / --enable-/--disable- autotools flags. Ogg
remains a hard dependency of Vorbis, Opus and Speex; ENABLE_EXTERNAL_LIBS
is kept as a backward-compat master switch.

Adapted from PR https://github.com/libsndfile/libsndfile/pull/400
This is a follow-up from
  https://github.com/libsndfile/libsndfile/pull/812#issuecomment-1088264645


commodo:
Well, a follow-up 4 years later... :p

arthurt (approved):
I'm very happy to see this happen. As the author of #400, I might not be an impartial reviewer though. @SoapGentoo ?

~~Could you expand more on why you left `HAVE_EXTERNAL_XIPH_LIBS`? I would think it would be better to get rid of it.~~ Nevermind, I see you mean you added a new master config flag.

commodo:
> I'm very happy to see this happen. As the author of #400, I might not be an impartial reviewer though. @SoapGentoo ?
> 
> ~Could you expand more on why you left `HAVE_EXTERNAL_XIPH_LIBS`? I would think it would be better to get rid of it.~ Nevermind, I see you mean you added a new master config flag.

I had to check again the code for that question about `HAVE_EXTERNAL_XIPH_LIBS`
I removed that.
The only "master" config flag now is HAVE_OGG, which disables Vorbis, Opus and Speex if not enabled (because they don't make sense without Ogg support).

PR #1130, opened 2026-06-14
ci: add permissions block and pin actions/checkout to full commit SHA

XananasX7:
`action.yml` was missing a top-level `permissions:` block and used a mutable action tag.

- Add `permissions: contents: read` (least privilege)
- Pin `actions/checkout`: `@v4` → `@34e114876b0b11c390a56381ad16ebd13914f8d5`

Qualifies under the [Google Open Source Patch Rewards Program](https://bughunters.google.com/open-source-security/patch-rewards).

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -2,6 +2,9 @@ name: C/C++ CI
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/action.yml
```diff
@@ -178,7 +181,7 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
erikd, line 184: Probably a bit overkill to use a commit hash, but all good.

erikd (approved):
LGTM.

PR #1140, opened 2026-07-12
wavlike: fix heap buffer over-read from unterminated cue label

saddamr3e:
1. the LIST/adtl `labl` handler in `wavlike_subchunk_parse` copies `sizeof(cue_points[i].name)` (256) bytes into the cue name with memcpy, so a label of 256+ bytes leaves the name with no NUL terminator.
2. downstream code then treats the name as a C string (aiff_write_header strlen's it when a WAV with cues is converted to AIFF), reading past the end of the SF_CUES allocation.

Switched to psf_strlcpy, matching the AIFF MARK reader in aiff.c. Added a command_test case that reads a crafted labl and checks the stored name stays terminated.


arthurt (approved):
LGTM

PR #1142, opened 2026-07-15
alac: fix pakt_size + 4 overflow in alac_pakt_read_decode.

hudsonzuo:
The pakt_size value is derived from chunk_info.datalen, which is read directly from the file chunk without any validation. In the vulnerable function, an addition (e.g., pakt_size + 5 or pakt_size + 4) is performed before memory allocation. When pakt_size is close to UINT32_MAX, this addition wraps around.


arthurt (approved):
Nice catch.

erikd:

PR #1143, opened 2026-07-16
ogg_opgus: fix memleak

hudsonzuo:
When malloc(oopus->buffersize) fails, it returns SFE_MALLOC_FAILED without freeing the oopus->buffer already allocated at L740. Repeated open failures can amplify the leak.


arthurt (approved):
Nice catch!

PR #1147, opened 2026-07-21
ircam: use psf_lrintf when reading samplerate

saddamr3e:
UBSan, clang -fsanitize=undefined:

    src/ircam.c:164:23: runtime error: 2.14748e+09 is outside the range of representable values of type 'int'

ircam_read_header reads the sample rate as a float straight from the file, then casts it with `(int) samplerate`. A crafted IRCAM header carrying an out-of-range or non-finite value makes that cast undefined. psf_lrintf gives defined behaviour and matches mat4.c/mat5.c, which read a file sample rate the same way; the out-of-range value is then rejected by validate_sfinfo. Found with the ossfuzz harness; all 143 ctest cases still pass.

### DIFF
--- a/src/ircam.c
+++ b/src/ircam.c
@@ -161,7 +161,7 @@ ircam_read_header	(SF_PRIVATE *psf)
 
 	psf_log_printf (psf, "marker: 0x%X\n", marker) ;
 
-	psf->sf.samplerate = (int) samplerate ;
+	psf->sf.samplerate = psf_lrintf (samplerate) ;
 
 	psf_log_printf (psf,	"  Sample Rate : %d\n"
 							"  Channels    : %d\n"


erikd (approved):
Seems sensible.

erikd:
Some Windows built issues.

erikd (approved):
Seems sensible.

saddamr3e:
Looked into it, and the Windows failures aren't from this patch. windows-latest now resolves to the windows-2025-vs2026 runner image, which dropped VS 2022, and the workflow invokes cmake with -G "Visual Studio 17 2022", so configure fails before any source is compiled. The latest master run fails the same four jobs: https://github.com/libsndfile/libsndfile/actions/runs/30117437760. Pinning those jobs to os: windows-2022 (or moving the generator to VS 2026) should clear it. Can send a separate PR for the workflow if that helps.

saddamr3e:
Those Windows failures aren't from this change. CMake dies at configure ('Visual Studio 17 2022' generator reports 'could not find any instance of Visual Studio') before anything is compiled, and the latest master push (the ogg_opus memleak commit) fails the same four jobs the same way. Looks like the windows-latest runner image dropped VS2022 while the workflow still hardcodes that generator.

erikd:
> Those Windows failures aren't from this change.

I agree. The WIndows builds seem to be failing due to changes in Github Actions. @evpobr Maybe you could look at this?

erikd:
> Looks like the windows-latest runner image dropped VS2022 while the workflow still hardcodes that generator.

@saddamr3e Can the VS generator be changed? I have no idea. I have not used Windows for nearly 2 decades.

saddamr3e:
> Can the VS generator be changed?

Yes. The new runner image ships VS 2026 and its cmake already knows the matching generator (added in cmake 4.1), so it's a plain string swap in .github/workflows/action.yml:

```
sed -i "s/Visual Studio 17 2022/Visual Studio 18 2026/; s/windows-vs2022/windows-vs2026/" .github/workflows/action.yml
```

(the second substitution just renames the four jobs to match). I had the commit ready to send as a PR but GitHub rejects pushes that touch workflow files because my token lacks the workflow scope, so it needs someone with write access to apply. The fallback is pinning those four jobs to os: windows-2022, which still carries VS 2022, but that only holds until that image gets retired too.

erikd:
Fingers crossed: https://github.com/libsndfile/libsndfile/pull/1148

erikd:
@saddamr3e The VS Generator PR passed CI and has been merged. Please rebase this one and force push.

saddamr3e:
Rebased and force pushed. One wrinkle: my token can't push the workflow commit to my fork (no workflow scope), so the branch is based on the commit just before the CI fix. Doesn't matter for the checks though, since PR runs take the workflow from the merge with master. The four Windows jobs are already running as windows-vs2026.

PR #1149, opened 2026-08-03
rf64, w64: ignore a second 'fmt ' chunk

saddamr3e:
wav.c already skips a second 'fmt ' chunk, but rf64.c and w64.c re-parse every one they see, and wavlike_read_fmt_chunk overwrites psf->sf.channels each time. The buffers sized from that count are allocated earlier and never revisited: wavlike_read_peak_chunk callocs peak_info for the channel count in effect when PEAK was read, and the WAVE_FORMAT_EXTENSIBLE branch does the same for channel_map. A file that puts a second 'fmt ' after those allocations with a larger channel count leaves the count and the buffers out of step, and the sf_command handlers size their access from the count rather than the allocation, so SFC_GET_MAX_ALL_CHANNELS and SFC_GET_CHANNEL_MAP_INFO hand adjacent heap back to the caller. I noticed it diffing the three wavlike header parsers against each other. ASAN on a stock build confirms both: ds64 + fmt(1ch) + PEAK + fmt(64ch) over-reads the 32 byte peak_info at command.c:397, and an EXTENSIBLE fmt(1ch) followed by fmt(64ch) reads 256 bytes past a 4 byte channel_map at sndfile.c:1431. Those two files are the blobs in the new test, which checks the channel count still comes from the first chunk. rf64 seeks past the skipped payload because its chunk loop doesn't, while w64's loop tail already does.


saddamr3e:
any update?

erikd:

Issue #1150, opened 2026-08-05
[Bug]misaligned 4-byte store in psf_binheader_readf while parsing a crafted WAV 'fmt ' chunk (CWE-704)


1820893135-pixel:
## Summary

When a WAV file's `wFormatTag` is `0x0055` (WAVE_FORMAT_MPEGLAYER3), `wavlike_read_fmt_chunk` calls `psf_binheader_readf` with the format string `"24222"`. Inside `psf_binheader_readf`'s `'4'` case, the code casts a destination pointer to `unsigned int*` and performs two 4-byte stores: `*intptr = 0` (common.c:1066) and `*intptr = psf_get_le32(...)` (common.c:1072). The target field lives in the library's internal `wav_fmt` union at an address that is **not 4-byte aligned** (observed `0x61600000009e`, where `0x9e % 4 == 2`). Writing an `unsigned int` through a misaligned pointer is undefined behavior per the C standard: it happens to work on x86, but on strict-alignment architectures (ARM/RISC-V/SPARC) it raises a bus error, and in the fail-fast build (`-fno-sanitize-recover=all`) UBSan traps via an illegal `ud1` instruction (**SIGILL**), terminating the process.

- **Affected versions**: libsndfile 1.2.2 / HEAD `7ff854d`
- **Severity**: Low (UB/misaligned access; DoS in sanitizer/strict-alignment builds)
- **CWE**: CWE-704 (Incorrect Type Conversion or Cast)
- **Sanitizer**: UndefinedBehaviorSanitizer

### Affected code

```c
/* src/common.c:1066 & 1072 (excerpt), in psf_binheader_readf's '4' case */
case 4:
    ...
    unsigned int *intptr = (unsigned int *)ptr;   /* <-- casts to unsigned int* */
    *intptr = 0;                                   /* common.c:1066 -- misaligned store */
    ...
    *intptr = psf_get_le32(psf, 0);                /* common.c:1072 -- misaligned store */
```

Call chain:

```
psf_binheader_readf (common.c:1066/1072, misaligned unsigned int store)
  └─ wavlike_read_fmt_chunk (wavlike.c:312, format string "24222" when wFormatTag==0x0055)
       └─ wav_read_header (wav.c:396)
            └─ wav_open (wav.c:174)
                 └─ psf_open_file (sndfile.c:3165)
                      └─ sf_open_virtual (sndfile.c:513)
                           └─ LLVMFuzzerTestOneInput (harness.c:146)
```

[...]

PR #1152, opened 2026-08-10
common: Use memcpy for unaligned field stores

4ktLuffy:
`psf_binheader_readf()` stores decoded header values through pointers obtained from `va_arg` and cast to a wider type. The destination fields live inside packed on-disk header structs and are not guaranteed to be naturally aligned, so these stores are undefined behaviour — fine on x86, but they fault on strict-alignment targets and UBSan traps them.

Reproducing #1150 with the 44-byte WAV from that issue (`wFormatTag` 0x0055, so `wavlike_read_fmt_chunk()` uses the `"24222"` format string), built with `-fsanitize=alignment`:

```
common.c:1066:6: runtime error: store to misaligned address 0x...be for type
'unsigned int', which requires 4 byte alignment
    #0 psf_binheader_readf common.c:1066
    #1 wavlike_read_fmt_chunk wavlike.c:312
```

#1150 reports this against the `'4'` case, but the same construct is used by every multi-byte specifier (`m`, `h`, `2`, `3`, `4`, `8`, `f`, `d`), so this converts them all rather than just the one path that happens to have a reproducer. `memcpy` imposes no alignment requirement and compiles to the same load/store on targets that allow unaligned access.

`'f'` and `'d'` also read directly into the unaligned destination; they now read into the existing scratch buffer first, as the other cases already do. The single-byte `'1'` case is unchanged since it cannot be misaligned.

### Testing

- The #1150 reproducer no longer trips UBSan under `-fsanitize=alignment -fno-sanitize-recover=alignment`.
- Behaviour on that file is unchanged: both before and after, it is rejected with `Malformed 'fmt ' chunk` — the fix removes the UB without changing parsing.
- Five valid WAVs (1/2 channel, 8/16/24/32-bit, 8 kHz–96 kHz) parse to identical results before and after.
- `ctest`: 143/143 pass.
- `src/common.c` compiles clean with `-Wall -Wextra`.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -979,11 +979,42 @@ header_gets (SF_PRIVATE *psf, char *ptr, int bufsize)
 	return k ;
 } /* header_gets */
 
+/*
+** Destination fields passed to psf_binheader_readf() live inside packed
+** on-disk header structs and are not guaranteed to be naturally aligned.
+** Storing through a cast pointer is undefined behaviour and faults on
+** strict-alignment targets, so route every multi-byte store through memcpy().
+*/
+static inline void
+store_uint (void *ptr, unsigned int value)
+{	memcpy (ptr, &value, sizeof (value)) ;
+} /* store_uint */
+
+static inline void
+store_ushort (void *ptr, unsigned short value)
+{	memcpy (ptr, &value, sizeof (value)) ;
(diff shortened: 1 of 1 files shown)


erikd:

PR #1154, opened 2026-08-15
Compute the header byte rate without overflowing int

manuaudio:
### What

`wav`, `w64` and `rf64` each write a bytes-per-second header field as

```c
BHW4 (psf->sf.samplerate * psf->bytewidth * psf->sf.channels)
```

All three operands are `int`, and the sample rate comes from the *input* file. Converting a file that declares an implausible rate overflows the multiply, which is undefined:

```
src/wav.c:800:38: runtime error: signed integer overflow:
1543547972 * 2 cannot be represented in type 'int'
    #0 wav_write_fmt_chunk wav.c:800
    #1 wav_write_header wav.c:1185
```

The same input reproduces it in `w64.c:479` and `rf64.c:548`.

Only the writer is affected — `sf_open` on the malformed file is fine — so it takes a conversion to reach, which `sndfile-convert` does. The reproducer is a PAF file declaring a sample rate of 1543547972.

### The change

`BHW4` is `((uint32_t) (x))`, so doing the arithmetic in `sf_count_t` and letting `BHW4` truncate writes **the same four bytes it always did**, while removing the overflow. That is also how `au.c`, `avr.c` and `caf.c` already guard the same shape.

Ten sites, all the identical expression: five in `wav.c`, four in `w64.c`, one in `rf64.c`.

### Verification

| | wav | w64 | rf64 |
|---|---|---|---|
| before | 2 runtime errors | 2 | 2 |
| after | 0 | 0 | 0 |

Reverting just the source change and rebuilding brings all six back, so the result is tied to the change and not to the input.

Ordinary files are unaffected: converting the same source to WAV before and after gives byte-identical output, sha1 `3a3816cf706e4e8bcf4959989b853be3a27dc88d`. Converting every format in my seed corpus to wav/w64/rf64 produces no sanitizer output.

### Two related spots I did not touch

- `sf_current_byterate` (`sndfile.c:1653`) has the same multiply but returns `int`, so what it *should* return for an implausible rate is a design question rather than a cast.
- `wavlike_srate2blocksize` takes an `int`, so its callers can overflow before the call.

Happy to fold either in if you'd like them handled together.

### How it was found

Fuzzing the command line tools — which the OSS-Fuzz harnesses don't cover, since those drive `sf_open`/read rather than `sndfile-convert` — with mutated real audio files, under ASan and UBSan.

I calibrated that setup before trusting it: with a deliberate one-byte over-read injected into `wav_read_header`, ASan reports it through `sndfile-info`, and the report disappears when the injection is removed. Worth mentioning that it did *not* reproduce the `%s` over-read fixed in #1141 — on this build the byte after the struct happens to be zero — so a clean run from it is not evidence about that class.

### Disclosure

Written with Claude Code assisting, under my direction; the commit carries an `Assisted-By` trailer. I verified the before/after myself on macOS/arm64. I did not find an AI policy in the repo — happy to follow one if you have a preference.

X: manuaudiomix

### DIFF
--- a/src/rf64.c
+++ b/src/rf64.c
@@ -545,7 +545,7 @@ rf64_write_fmt_chunk (SF_PRIVATE *psf)
 			/* fmt : format, channels, samplerate */
 			psf_binheader_writef (psf, "4224", BHW4 (fmt_size), BHW2 (WAVE_FORMAT_EXTENSIBLE), BHW2 (psf->sf.channels), BHW4 (psf->sf.samplerate)) ;
 			/*  fmt : bytespersec */
-			psf_binheader_writef (psf, "4", BHW4 (psf->sf.samplerate * psf->bytewidth * psf->sf.channels)) ;
+			psf_binheader_writef (psf, "4", BHW4 ((sf_count_t) psf->sf.samplerate * psf->bytewidth * psf->sf.channels)) ;
 			/*  fmt : blockalign, bitwidth */
 			psf_binheader_writef (psf, "22", BHW2 (psf->bytewidth * psf->sf.channels), BHW2 (psf->bytewidth * 8)) ;
 
--- a/src/w64.c
+++ b/src/w64.c
@@ -148,7 +148,7 @@ w64_open	(SF_PRIVATE *psf)
 		psf->blockwidth = (sf_count_t) psf->bytewidth * psf->sf.channels ;
 
 		if (subformat == SF_FORMAT_IMA_ADPCM || subformat == SF_FORMAT_MS_ADPCM)
-		{	blockalign = wavlike_srate2blocksize (psf->sf.samplerate * psf->sf.channels) ;
+		{	blockalign = wavlike_srate2blocksize ((sf_count_t) psf->sf.samplerate * psf->sf.channels) ;
 			framesperblock = -1 ;
 
 			/*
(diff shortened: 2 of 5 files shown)


manuaudio:
I need to correct something I wrote above. I said `wavlike_srate2blocksize` was left out because its callers overflow before the call — the reasoning was right, but I described it as untested when in fact my check simply never ran. I had used `-imaadpcm`, which `sndfile-convert` rejects as an unknown argument, and I read the absence of a report as absence of a bug.

With the correct `-ima-adpcm` / `-ms-adpcm` it reproduces from the same input:

```
src/wav.c:219:62: runtime error: signed integer overflow:
1543547972 * 2 cannot be represented in type 'int'
```

at `wav.c` 219, 849, 870 and `w64.c` 151, 534, 559. And the line right after each one,

```c
bytespersec = (psf->sf.samplerate * blockalign) / framesperblock ;
```

overflows the same way once `blockalign` is a few hundred — `wav.c` 851, 872 and `w64.c` 536, 561.

So I have pushed a second commit covering those. `wavlike_srate2blocksize` only compares its argument against 12000, 23000 and 44000, so widening the parameter to `sf_count_t` costs nothing and lets callers pass the product without wrapping. Its signature changes, which is why it is a separate commit — easy to drop if you would rather keep this PR to the first one.

Verification, same shape as before:

| | before | after |
|---|---|---|
| `-ima-adpcm` | 3 runtime errors | 0 |
| `-ms-adpcm` | 3 | 0 |
| default → wav/w64/rf64 | 2 each | 0 |

Encoding an ordinary file is byte-identical across the change: `-ima-adpcm` gives `d1a21ffdfdbaaeb8cb6f2eb38177977e7b8a92ce` and `-ms-adpcm` gives `e5ed4e22a02ba995114e39486877d29ba4a7a4e7` both before and after.

`sf_current_byterate` still stands apart — it returns `int`, so what it should do with an implausible rate is a design question rather than a cast, and I could not find a tool that reaches it with such a file.

erikd:

PR #1159, opened 2026-09-01
.github/workflows: Update GH Action references

erikd:
Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


erikd:
@copilot That PR only changes the FP precision on CI builds. Can that change be moved to the default CMake files so that all builds get this change?

Issue #1157, opened 2026-08-23
SIGFPE in vorbis_rshort()


stsp:
[test.tar.gz](https://github.com/user-attachments/files/31348090/test.tar.gz)
Attached is the fully automated test case.
Just run `make`.

The crash happens here:
```
Thread 12 "repro" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7ffff0fae6c0 (LWP 1888903)]
0x00007ffff7ccbd7e in vorbis_rshort (psf=0x7fffe40008f0, samples=576, 
    vptr=0x7fffe402fa60, off=0, channels=1, pcm=0x7fffe4010420)
    at /home/stas/src/libsndfile/src/ogg_vorbis.c:557
557			float inverse = 1.0 / psf->float_max ;
(gdb) p psf->float_max
$1 = 0
(gdb) bt
#0  0x00007ffff7ccbd7e in vorbis_rshort
    (psf=0x7fffe40008f0, samples=576, vptr=0x7fffe402fa60, off=0, channels=1, pcm=0x7fffe4010420) at /home/stas/src/libsndfile/src/ogg_vorbis.c:557
#1  0x00007ffff7ccc259 in vorbis_read_sample
    (psf=0x7fffe40008f0, ptr=0x7fffe402fa60, lens=3200, transfn=0x7ffff7ccbd1b <vorbis_rshort>) at /home/stas/src/libsndfile/src/ogg_vorbis.c:634
#2  0x00007ffff7ccc3fa in vorbis_read_s
    (psf=0x7fffe40008f0, ptr=0x7fffe402fa60, lens=3200)
    at /home/stas/src/libsndfile/src/ogg_vorbis.c:667
#3  0x00007ffff7c96aeb in sf_readf_short
    (sndfile=0x7fffe40008f0, ptr=0x7fffe402fa60, frames=3200)
    at /home/stas/src/libsndfile/src/sndfile.c:1836
```

PR #1163, opened 2026-09-06
ogg_vorbis: skip the float-to-int scaling when the signal peak is zer…

AetherAI3:
Fix for #1157.

### What's happening

`vorbis_rshort` and `vorbis_rint` in `src/ogg_vorbis.c` both compute

```c
float inverse = 1.0 / psf->float_max ;
```

whenever `psf->float_int_mult` is on. `psf->float_max` is populated
once inside the `SFC_SET_SCALE_FLOAT_INT_READ` handler, via
`psf_calc_signal_max`. When the peak-scan samples are all zero — a
decoded ogg-vorbis stream that holds nothing but silence — `float_max`
lands as `0`.

Under a caller that has FPU div-by-zero traps enabled (dosemu2's
openmt32 branch turns on `FE_DIVBYZERO` before it calls into
fluidsynth, which is how #1157 was reported), `1.0 / 0.0` raises
`SIGFPE`. In plain IEEE mode the result is `+Inf` and every produced
short is whatever `psf_lrintf` makes of `Inf * sample`.

### The fix

Guard both sites with `&& psf->float_max > 0.0`. When the peak is
zero the `else` branch runs, which multiplies each sample by
`32767.0f` directly. A zero sample stays zero, so the numerical
output is the same as before; the only difference is the pathological
input no longer divides by zero.

+2 / −2, `src/ogg_vorbis.c` only.

### How it was verified

- Wrote a small silent ogg two ways: `oggenc -q -1` on 2 ms of `sox`
  silence, and libsndfile itself writing all-zero floats through
  `SF_FORMAT_OGG | SF_FORMAT_VORBIS`.
- Instrumented `vorbis_rshort` on `master` at `b9103bd` to print
  `float_int_mult` and `float_max` — confirmed `float_int_mult=1,
  float_max=0` in both cases.
- A harness that enables `feenableexcept(FE_DIVBYZERO)`, opens the
  ogg, calls `sf_command(sf, SFC_SET_SCALE_FLOAT_INT_READ, &1,
  sizeof(int))`, then `sf_readf_short`. Baseline: SIGFPE, sigsetjmp
  handler catches signo 8. Fixed: OK read, exit 0.
- Regression: a real 0.1 s 440 Hz sine through the same harness still
  takes the scaled fast path and reads the full 800 frames — the
  guard only redirects the truly-silent case.

### Same shape elsewhere

The identical `if (psf->float_int_mult) { … 1.0 / psf->float_max …
}` shape lives in `src/ogg_opus.c` (two sites) and, with an added
default-1.0 ternary, in `src/float32.c` and `src/double64.c`. Kept
out of this patch on purpose — the report is specifically
`ogg_vorbis` and the fix here matches that scope. Happy to send
follow-ups if you'd like the same guard everywhere.

### Reporter

Thanks to @stsp for the report — the gdb trace inside the ticket
already names the exact line and the value of `float_max`, which is
what let this ship as a small, contained change.

### AI assistance

I used an AI assistant to help draft the fix, the reproduction and
this description. Every line was read, checked and run by me.

- [x] I have re-read every line and this description is in my own words.


stsp:
> Reported-by: Stanislav Brabec (github.com/stsp)

AI hallucinated the name.
But I don't care if the content is valid.

